PR #1, opened 2023-09-20
feat: add printenvcfg rule

avisiedo:
This new rule will help to know the configuration environment variables that are exported for the commands executed by the makefile.

### DIFF
--- a/scripts/mk/printvars.mk
+++ b/scripts/mk/printvars.mk
@@ -6,3 +6,7 @@
 .PHONY: printvars
 printvars: ## Print variable name and values
 	@true $(foreach V, $(sort $(.VARIABLES)),$(if $(filter-out environment% default automatic,$(origin $V)),$(info $V=$(value $V))))
+
+.PHONY: printenvcfg
+printenvcfg: ## Print the environment the resulting exported environment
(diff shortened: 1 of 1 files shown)


scripts/mk/printvars.mk
```diff
@@ -6,3 +6,7 @@
 .PHONY: printvars
 printvars: ## Print variable name and values
 	@true $(foreach V, $(sort $(.VARIABLES)),$(if $(filter-out environment% default automatic,$(origin $V)),$(info $V=$(value $V))))
+
+.PHONY: printenvcfg
+printenvcfg: ## Print the environment the resulting exported environment
+	@env | grep -e ^DATABASE_ -e ^KAFKA_ -e ^APP_ -e ^WEB_ -e ^LOGGING_ -e ^CONFIG_PATH= | sort
```
tiran, line 12: The grep pattern `^APP_` does not match the env var `APP`, because it does not end with an underscore. That's the env var that broke our configuration. Here is a better regular expression with Perl syntax:

```suggestion
	@env | grep -P -e '^(APP|CONFIG_PATH|DATABASE|KAFKA|LOGGING|WEB)(_.*)?=' | sort
```

tiran (changes requested):

tiran (approved):

PR #5, opened 2023-09-21
fix: Misc ephemeral deployment fixes

tiran:
`build_deploy.sh` runs container-push after container-build (fix for parallel builds with `MAKEFLAGS` env var)

`make ephemeral-deploy` creates secrets directory, `bonfire.yaml` from example, and fails if `secrets/private.mk` is missing.

`make ephemeral-build-deploy` fails if the `build_deploy.sh` script fails.

`ephe-domains-token.sh` uses correct XRHID identity type. Tokens are requested as user, not system.


avisiedo (approved):
Thanks for the changes:

* I have deployed in ephemeral by `make ephemeral-deploy` with no issues.
* I have created and registered a domain by using the ./test/scripts/ in the ephemeral environment successfully (using ORG_ID=12345).

LGTM

PR #7, opened 2023-09-21
fix: compose-up without configs/config.yaml

tiran:
The `db migrate-up` step of `make compose-up` no longer fails when `configs/config.yaml` does not exist or has different settings.

Also indent some lines to indicate that it's a continuation line.


avisiedo (approved):
LGTM

I have tried locally by: `make compose-clean clean build lint vet test compose-up run` and the db-tool side started successfully; and I can see the values are propagated when launching `db-migrate-up`.
I can see the below when running `make compose-up`
```raw
make DATABASE_USER="idmsvc" DATABASE_PASSWORD="idmsvc" DATABASE_NAME="idmsvc" DATABASE_HOST="localhost" DATABASE_PORT="5432" db-migrate-up
```

LGTM

tiran:
We should figure out how to automate the compose-up step.

PR #6, opened 2023-09-21
fix: add org_id to the options

avisiedo:
When we register a domain, we need to specify the org_id for the generated identity; the org_id we use here must match the org_id used when generating the token or the operation will fail with sign mismatching.

This change add a second option to specify the ORG_ID to be used for the identity.

PR #8, opened 2023-09-21
tests: Test scripts lookup ORG_ID in env

tiran:
Tests scripts in `test/scripts` were using a hard-coded organization id for XRHID header. The scripts are now looking up `ORG_ID` in environment and fall back to `12345` test org.

The improvement allows testing with different organization ids.


avisiedo:
thank you; I appreciate this change, and the suggestions are incorporated at https://github.com/podengo-project/idmsvc-backend/pull/6 PR which update a bigger change into the set of scripts.

I close this ticket.

PR #10, opened 2023-09-28
HMS-2595 fix: GET signing_keys dummy

tiran:
The GET `signing_keys` route now returns an empty result set instead of an error. The change is necessary to unbreak integration testing of`ipa-hcc`.


avisiedo (approved):
LGTM

PR #11, opened 2023-09-28
HMS-2590 fix: use 'resources' tag for PATCH /domains/:domain_id

avisiedo:
Update the generated spec.gen.go file after the new API has been attached.

### DIFF
--- a/internal/api/public/spec.gen.go
+++ b/internal/api/public/spec.gen.go
@@ -108,14 +108,14 @@ var swaggerSpec = []string{
 	"a27K41WvdWfc/pQkBC0ll3mYVcRhJ+J38abLFxIz7j8pxstaUZuZAblQZgSMMx1Ucbq6zCV6ZNWbzvKC",
 	"4yOMgQQYtf8yhyoJuU5LGdnU9EppRssYcCe6fBqi2RiiVvYIcKQoBIjIn1U606kyEXflNCe98s1s7ikG",
 	"SbTW2qfvgFvbuHAVz4j/Rc6XQ97ACj2fiVjv+RT/0IawKZA58zIfyGyXlDGVfctdvYwtyodtZsm+k12p",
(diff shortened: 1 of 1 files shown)


tiran:
Could you please change the commit message and PR to reflect the details of the change, while focusing on relevant changes? The fact only that `spec.gen.go` has changed, is not interesting. As a reviewer and developer, I'm interested to know what has in the file.

Also use upper-case letters for title reference

```
HMS-2590 fix: Regenerate OpenAPI for PATCH /domains/:domain_id

Update to latest OpenAPI spec. `PATCH /domains/:domain_id` now has `resources` tag instead of `actions` tag.
```

avisiedo:
@tiran thanks mate! updated!

tiran:
You updated the ticket description, but not the commit message that will land in git. The commit message is IMHO more important than the ticket description. After all the information ends up in the commit history.

A sentence or two is more than sufficient for a trivial change. The change doesn't even affect the backend.

avisiedo:
@tiran  thanks mate! I didn't realize about the commit message; now it is updated.

tiran:
I hate do be pedantic ... but you didn't read my last comment, did you? :) Now the commit message is overly long and goes into unimportant details.

In my opinion and experience, commit messages should provide a short overview of the "what" and "why" of the change set. They should be as long as necessary and as short as possible. For this PR, one sentence, maybe two are more than sufficient. Personally I would not even include an explanation, because the change does not have an impact on the behavior of the backend. It only updates the serialized JSON string of the schema.

avisiedo:
> I hate do be pedantic ... but you didn't read my last comment, did you? :) Now the commit message is overly long and goes into unimportant details.
> 
> In my opinion and experience, commit messages should provide a short overview of the "what" and "why" of the change set. They should be as long as necessary and as short as possible. For this PR, one sentence, maybe two are more than sufficient. Personally I would not even include an explanation, because the change does not have an impact on the behavior of the backend. It only updates the serialized JSON string of the schema.

thanks mate! at the beginning I understood it was the PR message; later I updated the commit message; I have updated with a short description. :+1:

frasertweedale (approved):

PR #12, opened 2023-10-04
HMS-2692 feat: App secrets with HKDF

tiran:
Change internal secrets handling to derive domain secrets from a common app secrets with HKDF.

`app.domain_reg_key` has been removed. Instead the domain registration token key is derived from `app.secret` using RFC 5869 HKDF extract and expand. In the future, other secrets like JWK encryption key will be derived the same way.


frasertweedale:
LGTM.  Merging.  Thanks!

frasertweedale (approved):

PR #13, opened 2023-10-04
HMS-2695 feat: JWK encryption with AES-GCM

tiran:
Implement helpers to encrypt and decrypt a JWK with AES-GCM AEAD.

`EncryptJWK()` takes a raw AES encryption key plus `jwk.Key` and returns the encrypted key. The key is encrypted and protected with AES-GCM. The random nonce is pre-pended to the output stream.

`DecryptJWK()` is the reverse operation. It takes a raw AES key and cipher text with nonce pre-pended. On success, it returns a `jwk.Key`. Invalid key and tampering are detected by AEAD.

Also pulls the latest `lestrrat-go/jwx@v2` go module.


internal/infrastructure/token/encrypt.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	// serialize JWK to JSON bytes
+	jwkbuf, err := json.Marshal(jwkey)
+	if err != nil {
+		return nil, err
+	}
+
+	// encrypt and seal
+	ciphertext := aesgcm.Seal(nil, nonce, jwkbuf, nil)
```
frasertweedale, line 38: This Go crypto API is a bit odd, and not adequately documented.  From reading the source code, I *think* the follow would work too.  (Not that I am asking for this change; I prefer it explicit, the way you wrote it):

```go
    return aesgcm.Seal(nonce, nonce, jwkbuf, nil)
```

The first argument, `dst`, will be copied to the output.  In fact, if `cap(dst) >= len(dst) + len(plaintext) + tagSize` it will be used as-is, and the ciphertext and tag will simply be appended without making a new buffer.  So if you make the nonce buffer big enough and only read `NonceSize()` random bytes into it, there will only be one new buffer allocated for the whole encryption operation and the output is exactly what we want.

Again all of this is overkill for us, and harms readability.  I just want to share what I have learned about this API :)

internal/infrastructure/token/encrypt_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"testing"
+
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestJWKEncryption(t *testing.T) {
```
frasertweedale, line 13: Suggestion (nice to have, not a show stopper) - could we explicitly assert nonce uniqueness?  I think it would suffice to ensure that two encryptions of the same plaintext with same key result in different ciphertext.  (Because the only thing that *could* result in different ciphertext in that scenario is a different nonce.)

frasertweedale (changes requested):
One suggestion to extend the tests.

internal/infrastructure/token/encrypt_test.go
tiran: Sure, done!

I have encrypted the value twice with the same key and verified that the output is different.

frasertweedale (approved):

internal/infrastructure/token/encrypt_test.go
frasertweedale: Cheers mate!

PR #16, opened 2023-10-06
HMS-2694 feat: Host conf token signing

tiran:
Add helpers to build a host configuration token and to sign a token with multiple keys.

`BuildHostconfToken()` builds a new token and returns a token instance.

`SignToken()` signs a token with multiple keys and returns a JSON string.


frasertweedale (approved):
LGTM.  Thanks!

PR #17, opened 2023-10-09
HMS-2694 feat: Host conf token MVP

tiran:
Implement minimal host conf token and signing JWK.

A private ECDSA key is genenerated from a hard-coded seed and the JWK is
stored in memory. In the future, the ECDSA key will be generated from a
random source and the JWK stored in the database.

The `GET /signing_keys` endpoint returns the public JWK.

The `POST /host-conf` endpoint now returns a signed token.

Signed-off-by: Christian Heimes <cheimes@redhat.com>

### DIFF
--- a/internal/handler/impl/application.go
+++ b/internal/handler/impl/application.go
@@ -4,12 +4,16 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"hash"
 	"io"
+	"time"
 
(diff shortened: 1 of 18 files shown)


frasertweedale:
Diff looks good.  I'm going to do some testing before I merge it.

frasertweedale:
It works.  There are some things that need addressing but not in this diff.  Better to follow-up with subsequent PRs and keep this already-big PR from growing more.

So, I am merging it.  Thanks!

frasertweedale (approved):

PR #18, opened 2023-10-11
HMS-2694 fix: exclude disabled domains from host-conf

frasertweedale:
A domain with `auto_enrollment_enabled = FALSE` should not be
returned in a host-conf response.  Iterate the returned records (no
longer limited to 2) and exclude disabled domains.

Also improve the HTTP response status for some expected ways
host-conf can fail.

- No matching domains (500 -> 404)
- Multiple matching domains (500 -> 409)

Also remove a seemingly obsolete comment.

### DIFF
--- a/internal/usecase/repository/host_repository.go
+++ b/internal/usecase/repository/host_repository.go
@@ -1,10 +1,11 @@
 package repository
 
 import (
-	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/openlyinc/pointy"
 	"github.com/podengo-project/idmsvc-backend/internal/api/public"
 	"github.com/podengo-project/idmsvc-backend/internal/domain/model"
 	internal_errors "github.com/podengo-project/idmsvc-backend/internal/errors"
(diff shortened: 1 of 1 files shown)


internal/usecase/repository/host_repository.go
```diff
@@ -31,12 +33,12 @@ func (r *hostRepository) MatchDomain(db *gorm.DB, options *interactor.HostConfOp
...
 	}
 
-	// look through domains and find domains with non-NULL token
 	// TODO: match FQDN with domain realms
 	var domains []model.Domain
 	tx := db.Model(&model.Domain{}).
 		Joins("left join ipas on domains.id = ipas.id").
-		Where("domains.org_id = ?", options.OrgId)
+		Where("domains.org_id = ?", options.OrgId).
+		Where("domains.auto_enrollment_enabled = TRUE")
```
tiran, line 41: This change will result in the same error message when a domain is not found and a domain is disabled. For debugging and logging, it would be better to have different error results for "no domain found" and "domain is disabled".

frasertweedale: OK, I'll change the approach, but it will involve no longer limiting the number of records in the response set.  I think that's OK, we expect most orgs to have a small number, and we don't expect any to have a huge number of registered services.

frasertweedale: Besides, when it comes to "semantic" matching of the host domain name to choose the best candidate domain, we will have to iterate a set of records.  Because SQL is probably not expressive enough for what we would do.

frasertweedale:
converting to draft.

internal/usecase/repository/host_repository.go
tiran: We have not designed rules to match a FQDN to a domain. We'll probably need another table and additional APIs to create matching rules.

frasertweedale: I think we can hardcode matching rules with whatever heuristic(s) we think make sense.  And when customers beg us for something more flexible or different, *then* we work out whether we should do it, and if so, how :)

tiran (approved):

PR #20, opened 2023-10-17
fix: respond 404 when asked to delete unknown domain

frasertweedale:
Currently when deleting an unknown record the service responds with 500 Internal Server Error.  Discriminate this case and respond with 404 Not Found instead.


tiran (approved):

avisiedo (approved):
LGTM

PR #21, opened 2023-10-18
HMS-2714 fix description and title validation

frasertweedale:
The validation regexes for the description and title fields are too
restrictive.  Reasonable values are rejected.

Instead, use callback validators to check that the string values
(which are byte sequences) are valid UTF-8, and they do not contain
control characters (except the description field allows Carriage
Return, Line Feed, and Horizontal Tab).

ALSO

Currently input validation failure does NOT cause processing to
stop.  If somehow the backend operation succeed, that can result in
a 400 Bad Request response, but the target backend operation
completed and database modifications potentially performed.
Also, the normal response data gets appended to the error message.

Update the validation middleware to stop processing when an input
validation error occurs, after setting the response status and body
content (error message).  As a result, no further processing will be
performed.

### DIFF
--- a/internal/infrastructure/middleware/validate_api.go
+++ b/internal/infrastructure/middleware/validate_api.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/getkin/kin-openapi/openapi3filter"
@@ -46,8 +48,8 @@ func InitOpenAPIFormats() {
 	openapi3.DefineStringFormatCallback("cert-subject", func(value string) error {
 		return checkFormatSubject(value)
 	})
-	openapi3.DefineStringFormat("domain-description", `^[\n\x20-\x7E]*$`)
-	openapi3.DefineStringFormat("domain-title", `^[a-zA-Z0-9\s]+$`)
+	openapi3.DefineStringFormatCallback("domain-description", checkUtf8MultiLine)
+	openapi3.DefineStringFormatCallback("domain-title", checkUtf8SingleLine)
 	openapi3.DefineStringFormatCallback("ipa-realm-domains", func(value string) error {
 		return checkFormatRealmDomains(value)
 	})
@@ -63,6 +65,34 @@ func helperCheckRegEx(regex string, fieldName string, fieldValue string) error {
 	}
 	if !match {
 		return fmt.Errorf("'%s'='%s' format not matching", fieldName, fieldValue)
+	}
+	return nil
+}
+
+// Check that input is a valid UTF-8 string with no control characters
+// (not even CR/LF).
+func checkUtf8SingleLine(s string) error {
+	if !utf8.ValidString(s) {
+		return fmt.Errorf("not a valid utf-8 string")
+	}
+	for _ /* index */, r /* rune */ := range s {
+		if unicode.IsControl(r) {
+			return fmt.Errorf("invalid code point: %U", r)
+		}
+	}
+	return nil
+}
+
+// Check that input is a valid UTF-8 string with no control characters,
+// except spacing chars including CR/LF are allowed.
+func checkUtf8MultiLine(s string) error {
+	if !utf8.ValidString(s) {
+		return fmt.Errorf("not a valid utf-8 string")
+	}
+	for _ /* index */, r /* rune */ := range s {
+		if unicode.IsControl(r) && !unicode.IsSpace(r) {
+			return fmt.Errorf("invalid code point: %U", r)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


tiran (changes requested):
Code looks good to me.

Please add some positive and negative test cases to `validate_api_test.go`.

avisiedo:
I have checked the code by using the payload below for `test/data/http/patch-rhel-idm-domain.json` to force the failure on validation; See below the response is a plain text instead of json document; please see notes at the end.

```json
{
    "title": "mypatched.example \u0010",
    "description": "My patched description for mydomain.example domain",
    "auto_enrollment_enabled": true
}
```

To force the validation failure; on this situation the response I got is:

```raw
# ./test/scripts/local-domains-patch.sh "$UUID"
HTTP/1.1 400 Bad Request
Content-Type: text/plain
X-Frame-Options: SAMEORIGIN
X-Rh-Insights-Request-Id: test_4795483c7630454cbc97
Date: Wed, 18 Oct 2023 08:42:38 GMT
Content-Length: 139

request body has an error: doesn't match schema #/components/schemas/UpdateDomainUserRequest: Error at "/title": invalid code point:
```

Note `Content-Type: text/plain`; I guess it is coming from here: https://github.com/podengo-project/idmsvc-backend/blob/main/internal/infrastructure/middleware/validate_api.go#L233 

Note the response is not a json document, I guess is the line below: https://github.com/podengo-project/idmsvc-backend/blob/main/internal/infrastructure/middleware/validate_api.go#L234

----

I suggest returning by something similar to the below:

```golang
return echo.NewHTTPError(http.StatusBadRequest, err.Error())
```

----

**I am not sure about the same change at**:
* https://github.com/podengo-project/idmsvc-backend/blob/main/internal/infrastructure/middleware/validate_api.go#L257
* https://github.com/podengo-project/idmsvc-backend/blob/main/internal/infrastructure/middleware/validate_api.go#L258

----

Echo framework will handle the error response by the default error handler (https://echo.labstack.com/docs/error-handling#default-http-error-handler).

cryptomilk:
There are actually two places which need fixing:

```
internal/infrastructure/middleware/validate_api.go | 45 ++++++++++------------
 1 file changed, 20 insertions(+), 25 deletions(-)
 
diff --git a/internal/infrastructure/middleware/validate_api.go b/internal/infrastructure/middleware/validate_api.go
index 60d61a2..87354c9 100644
--- a/internal/infrastructure/middleware/validate_api.go
+++ b/internal/infrastructure/middleware/validate_api.go
@@ -260,9 +260,7 @@ func RequestResponseValidatorWithConfig(config *RequestResponseValidatorConfig)
 					requestValidationInput,
 				)
 				if err != nil {
-					c.Response().Header().Set(echo.HeaderContentType, "text/plain")
-					c.String(http.StatusBadRequest, err.Error())
-					return nil // stop processing
+					return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 				}
 			}
 
@@ -272,32 +270,29 @@ func RequestResponseValidatorWithConfig(config *RequestResponseValidatorConfig)
 				resRec := &ResponseRecorder{buffer: &bytes.Buffer{}, original: rw}
 				c.Response().Writer = resRec
 
-				defer func() {
-					responseValidationInput := &openapi3filter.ResponseValidationInput{
-						RequestValidationInput: requestValidationInput,
-						Status:                 resRec.status,
-						Header:                 resRec.Header(),
-					}
-					responseValidationInput.SetBodyBytes(resRec.buffer.Bytes())
-
-					if err = openapi3filter.ValidateResponse(
-						ctx,
-						responseValidationInput,
-					); err != nil {
-						// write error response
-						c.Response().Header().Set(echo.HeaderContentType, "text/plain")
-						c.String(http.StatusInternalServerError, err.Error())
-					} else {
-						// Write original response
-						rw.WriteHeader(resRec.status)
-						resRec.buffer.WriteTo(rw)
-					}
-				}()
-
 				defer func() {
 					// reset the response, using the original ResponseWriter
 					c.SetResponse(echo.NewResponse(rw, c.Echo()))
 				}()
+				err = next(c)
+				responseValidationInput := &openapi3filter.ResponseValidationInput{
+					RequestValidationInput: requestValidationInput,
+					Status:                 resRec.status,
+					Header:                 resRec.Header(),
+				}
+				responseValidationInput.SetBodyBytes(resRec.buffer.Bytes())
+
+				if err = openapi3filter.ValidateResponse(
+					ctx,
+					responseValidationInput,
+				); err != nil {
+					// write error response
+					return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+				}
+				// Write original response
+				rw.WriteHeader(resRec.status)
+				resRec.buffer.WriteTo(rw)
+				return err
 			}
 
 			return next(c)
```

tiran:
@cryptomilk Can you open a new PR with these fixes, please?

frasertweedale:
Thanks for the reviews fellas.  I'll update this PR by end of week.

tiran:
Testing our new bot

tiran (approved):

tiran:
LGTM, thanks for the tests.

avisiedo (approved):

frasertweedale:
Thanks @avisiedo and @tiran  - and @cryptomilk for filing follow-up PR #23

PR #24, opened 2023-10-20
HMS-2878 fix: Update go modules fixing security issues

cryptomilk:
Pair-Programmed-With: Andreas Schneider


tiran (changes requested):
- There is a typo in the commit message title.
- Please include a reference to the security issue (GHA#, CVE#, URL).

tiran (approved):

PR #25, opened 2023-10-23
infra: Enable Dependabot and merge tools go.mod

tiran:
Add `dependabot.yml` to enable Dependabot for GitHub Actions and Go modules.

Merge the `go.mod` and `go.sum` from the tools directory with `go.mod` in the root directory. This ensures that Dependabot will keep the dependencies in sync.


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
avisiedo, line 11:
```suggestion
  - package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "weekly"
  - package-ecosystem: "gomod"
    directory: "/tools"
    schedule:
      interval: "weekly"
```

avisiedo: suggestion: Would work the above? if yes, we could keep the tools dependencies; why to remove them?

avisiedo (commented):
I suggest a change that could avoid to remove **tools/go.mod**, and let dependabot to run. If that works, why to remove **tools/go.mod**?

tiran:
Is there any benefit in maintaining two different sets of  go module version? We are already jumping through some hoops to use the same version of oapi-codegen for tools and in the service.

avisiedo:
LGTM

sharing thoughts even if I were wrong, so please correct me:

(1) In the past I got some situations managing the dependencies in go at idmocp; at podengo-project when I realized that by adding **tools/** files that it was possible to keep the dependencies of the tools separate from the dependencies of the service, I thought it could be beneficial to avoid that the dependencies of the tools could impact into the service dependencies and generate situations as I had in the past. About **oapi-codegen**, I have realized recently that the tools does not depends on itself because the runtime that is used by the generated code is in a different package (even repository at https://github.com/oapi-codegen/runtime  the name is confusing as change the organization).

(2) Another reason it is future optimizations in the pipeline, so the tools could be pre-built in a container image or cached in the pipeline depending on tools/go.mod and tools/go.sum files; so dependency updates on the service would not evoke dependency updates on the tools, and reusing the pre-built binaries (from cache or container).

Said the above, nor (1) nor (2) are critical reasons to avoid the PR to merge, and we can come back to the above (1) at any moment in the future, and IMO use dependabot tool provides a high value to keep the dependencies in good shape.

> One last thing; dependabot could suggest in the MR to upgrade some indirect dependencies; if we merge that without detect the primary dependency could evoke situations when building.
> It could happens the primary dependency has no a release at the moment dependabot detect to bump the indirect dependency.

avisiedo (approved):
I pressed the wrong button! LGTM

PR #26, opened 2023-10-23
fix: Changing expiration time to never

aadhikar:
The images generated by the pipeline are deleted after 1d. This change fixes
the situation by setting the quay.expire-after label into the container images
by default to the value of 'never' which avoids the image being deleted. To let
the developer build its own images during the development lifecycle, that
value can be customized from the scripts/mk/private.example.mk template.

### DIFF
--- a/scripts/mk/container.mk
+++ b/scripts/mk/container.mk
@@ -47,7 +47,7 @@ registry-login:
 	$(CONTAINER_ENGINE) login -u "$(CONTAINER_REGISTRY_USER)" -p "$(CONTAINER_REGISTRY_TOKEN)" $(CONTAINER_REGISTRY)
 
 .PHONY: container-build
-container-build: QUAY_EXPIRATION ?= 1d
+container-build: QUAY_EXPIRATION ?= never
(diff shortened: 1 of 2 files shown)


avisiedo (changes requested):
Nice catch :+1:  I suggest to add the below too, to the `scripts/mk/private.example.mk` file.

```makefile
# QUAY_EXPIRATION is used to set the value for the 'quay.expires-after' label,   
# when the container image is built, so quay.io self-delete the image after the  
# indicated time. The value of `never` avoid the image is self-deleted.          
# https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/use_red_hat_quay/working_with_tags#setting_tag_expiration_from_a_dockerfile
export QUAY_EXPIRATION ?= 7d
```

The above file is used only by the developer as a once task when setting the development environment, so it would keep for development propose to self-delete feature for the development images pushed to its own registry during the software lifecycle. It would not affect to the value for the image built in the pipeline.

avisiedo (approved):

PR #33, opened 2023-10-25
infra: Use go.sum from go mod tidy

tiran:
`go get -d` and `go mod tidy` generate slightly different `go.sum` files. Dependabot uses `go mod tidy`, resulting in CI errors. Replace `go get -d` with `go mod download` and use the `go.sum` output of `go mod tidy`.


avisiedo (approved):
LGTM

PR #34, opened 2023-10-25
infra: Regenerate API and mock from Dependabot

tiran:
Use GHA to regenerate API and mock files, then rewrite Dependabot's commit.


tiran:
Does not work as expected. Right now it's easier to manually update the dependencies.

PR #40, opened 2023-11-02
compose: use correct account for healthcheck

frasertweedale:
The compose currently attempts to connect as the container user "root".  This role does not exist in the postgresql server.  The liveness test does succeed, but it fills the "log" (stdout) with annoying messages:

    ... FATAL:  role "root" does not exist

`pg_isready` recognises the standard *libpq* environment variables: https://www.postgresql.org/docs/16/libpq-envars.html.  Set the `PGDATABASE`, `PGUSER` and `PGPASSWORD` environment variables in the database container environment to ensure that the connect test uses a valid account.  As a result, the log output no longer contains the the above message every few seconds.

### DIFF
--- a/deployments/backend-compose.yaml
+++ b/deployments/backend-compose.yaml
@@ -1,3 +1,4 @@
+### This file is used by ipa-hcc CI jobs
 ---
 version: "3.0"
 services:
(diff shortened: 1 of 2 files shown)


tiran:
Could you please fix `deployments/backend-compose.yaml`, too?

frasertweedale:
Actually, I needed to ask about that file.  It is not referenced anyway.  I will take the question to the slack channel.

frasertweedale:
It's used by CI for ipa-hcc - I have updated it as requested.

tiran (approved):

PR #41, opened 2023-11-07
HMS-1832: refactor: PUT /domains/:domain_id using upserts (defered)

avisiedo:
(defered)

This change try to enhance how the PUT /domains/:domain_id endpoints works with the database layer, so to enhance the operations with the database, by using upserts and deleting only when necessary the remaining records for certificates, servers and locations.

TODO
- [x] Rebase

### DIFF
--- a/internal/usecase/repository/domain_repository.go
+++ b/internal/usecase/repository/domain_repository.go
@@ -122,29 +122,23 @@ func (r *domainRepository) Register(
 // 	return *data, nil
 // }
 
-// UpdateAgent save the Domain record into the database. It only update
-// data for the current organization.
+// UpdateAgent save the Domain record into the database.  The org_id
+// and domain_uuid from the data object are used to select the target
+// record.
+//
+// It is assumed that the record is complete and changes have already been
+// "merged" and are ready to save to the database.
 func (r *domainRepository) UpdateAgent(
 	db *gorm.DB,
 	orgID string,
 	data *model.Domain,
 ) (err error) {
-	var currentDomain *model.Domain
 	if err = r.checkCommonAndData(db, orgID, data); err != nil {
 		return err
 	}
 
-	// Check the entity exists
-	if currentDomain, err = r.FindByID(
-		db,
-		orgID,
-		data.DomainUuid,
-	); err != nil {
-		return err
-	}
-
-	if err = db.Omit(clause.Associations).
-		Where("org_id = ? AND domain_uuid = ?", orgID, currentDomain.DomainUuid).
+	if err = db.
+		Omit(clause.Associations).
 		Updates(data).
 		Error; err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


test/data/http/update-rhel-idm-domain.json
```diff
@@ -2,20 +2,20 @@
   "domain_name": "mydomain.example",
   "domain_type": "rhel-idm",
   "rhel-idm": {
-    "realm_name": "MYDOMAIN.EXAMPLE",
+    "realm_name": "UPDATE-MYDOMAIN.EXAMPLE",
```
tiran, line 5: It is not possible to modify the Kerberos realm or domain name of an IdM deployment. We should treat a change as an error. There are probably some other fields that we should treat as read-only after initial registration. Can you open a ticket for the problem?

internal/usecase/repository/domain_repository.go
```diff
@@ -407,50 +420,68 @@ func (r *domainRepository) createIpaDomain(
...
-	if db == nil {
-		return internal_errors.NilArgError("db")
-	}
-	if data == nil {
-		return internal_errors.NilArgError("data' of type '*model.Ipa")
-	}
-	if err = db.Unscoped().
-		Delete(data).Error; err != nil {
-		return err
+func (r *domainRepository) ipaFillUpdateAgentReminders(oldData *model.Domain, newData *model.Domain) (remCerts []uint, remServers []uint, remLocations []uint) {
```
tiran, line 423: Do you mean `remainder` (rest, residual) instead of `reminder` (recall, memory)?

internal/usecase/repository/domain_repository.go
```diff
@@ -407,50 +420,68 @@ func (r *domainRepository) createIpaDomain(
...
+	remLocations = nil
+
+	// Check cert reminders
+	OldLenCerts := len(oldData.IpaDomain.CaCerts)
+	NewLenCerts := len(newData.IpaDomain.CaCerts)
+	if OldLenCerts > NewLenCerts {
+		remCerts = make([]uint, OldLenCerts-NewLenCerts)
+		for i := NewLenCerts; i < OldLenCerts; i++ {
+			remCerts[i-NewLenCerts] = oldData.IpaDomain.CaCerts[i].Model.ID
+		}
```
tiran, line 435: I'm not sure whether I understand the algorithm to perform upserts. Does a new cert + removal of an existing cert recycle a database entry? That sounds wrong to me.

Would it help if `ipa-hcc` would include a persistent, unique id that identifies an entry (server, location, cert)? That would allow us match LDAP entries to database entries.

test/data/http/update-rhel-idm-domain.json
avisiedo: thanks mate! I have created a ticket for it!

internal/usecase/repository/domain_repository.go
avisiedo: yeah! thank you! I mean `remainder`  :+1:

internal/usecase/repository/domain_repository.go
avisiedo: About "Does a new cert + removal of an existing cert recycle a database entry?", yes that would be what is doing with this change; the data would be modified, but the internal database key is kept; below what I have observed:

----

- If I add Cert A + remove Cert B,
- the record that store Cert A information after the operation will hold the information of Cert B and the key of the record is kept.
- The DB has no entities that point out to Certs (the certs has a foreign key to the Ipa record hat own them).

> If the DB schema is extended in the future and some entity reference to the cert, then this operation as it is implemented on this PR would be wrong.

----

About "Would it help if ipa-hcc would include a persistent, unique id that identifies an entry (server, location, cert)?", I am not sure; if we update (without upserts) the data, we could split the record to update, the records to insert and the records to delete; for domains with a lot of information we could group on bulk data.

----

Would we prefer then the navy implementation? (I am not against that, as I think this could be an early optimization that maybe we want to delay for later).

avisiedo:
Setting this change as a draft but not closing, to defer the changes in a later time.

avisiedo:
closing ticket because staleness

PR #42, opened 2023-11-07
HMS-2056 test: initial smoke test components

avisiedo:
Initial set of components to create smoke tests for the backend service, and potentially similar approach could be useful for integration tests.

Requirements:
- When running from the workstation, you have executed previously `make compose-up`.
- When running from the pipeline, the required services are up and running.

TODO
- [x] For smoke tests it seems ok to directly provide a REST API structure, but maybe not for integration tests where we want to cover the error responses, even bad formed json documents; review and rethink about it.
- [x] Some smoke tests could wrap the current SuiteBase to register a domain (implies call to retrieve a token + call to register the domain), so it is available for the operation to cover; the **preparation** requests would happen at **SetupTest** method, so the smoke test to add focus in the specific operation only.
- [x] Consideration for parallel tests; use a random organization id, or implement a generator for organization id, so that every SuiteTest use a different data partition and it is avoided side effects between the different tests; **concern** what is the granularity when the tests are parallelized by golang? by **Test...** method, by **TestSuite**? **TODO address in a new ticket with low priority**. TODO at HMS-3167
- [x] Review developer documentation that fit the implementation used as reference.
- [x] Refactor the starting side of the code to reuse by smoke and integration tests and remove duplication; **think about which package will hold the code, bearing in mind the dependencies to and from it**.
- [x] Create an initial package `internal/builder` which will generate random data, but allowing customization. **this will be a progressive task, but it is provided an initial extensive set of builders with methods for the customization**.

Notes:
- ListDomain test will be finished on HMS-3166
- UpdateDomain test will be implemented on HMS-3156
- ReadDomain test will be implemented on HMS-3154
- PatchDomain test will be implemented on HMS-3155
- Parallel Smoke Tests will be implemented on HMS-3167

**Update**: The last change to the PR includes HMS-3166 HMS-3156 HMS-3154 HMS-3155 and HMS-3167 so link that to close the pr when this is merged.

- https://github.com/podengo-project/idmsvc-backend/pull/60
- https://github.com/podengo-project/idmsvc-backend/pull/62
- https://github.com/podengo-project/idmsvc-backend/pull/63
- https://github.com/podengo-project/idmsvc-backend/pull/70
- https://github.com/podengo-project/idmsvc-backend/pull/71

Close #60 #62 #63 #70 #71

### DIFF
--- a/internal/test/builder/helper/helper.go
+++ b/internal/test/builder/helper/helper.go
@@ -1,0 +1,296 @@
+package helper
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/openlyinc/pointy"
+	"github.com/pioz/faker"
+)
+
+// GenRandNum generate a random number >= min and <= max interval.
+// min is the lower boundary interval including the value.
+// max is the higher boundary interval including the value.
+func GenRandNum(min, max int64) int64 {
+	return faker.Int64InRange(min, max+1)
+}
+
+// GenPastNearTime generate a past timestamp not further than delta.
+// delta is the duration that set the threshold for the time.
+// Return a time.Time for the expected interval.
+func GenPastNearTime(delta time.Duration) time.Time {
+	var value int64 = GenRandNum(0, int64(delta))
+	return time.Now().UTC().Add(time.Duration(value) * -1)
+}
+
+// GenFutureNearTimeUTC generate a past timestamp not further than delta.
+// delta is the duration that set the threshold for the time.
+// Return a time.Time for the expected interval.
+func GenFutureNearTimeUTC(delta time.Duration) time.Time {
+	var value int64 = GenRandNum(0, int64(delta))
+	return time.Now().UTC().Add(time.Duration(value))
+}
+
+// GenBetweenTimeUTC generate a timestamp between the given parameters
+// as earlier as 'begin' and before 'end'.
+// delta is the duration that set the threshold for the time.
+// Return a time.Time for the expected interval.
+func GenBetweenTimeUTC(begin time.Time, end time.Time) time.Time {
+	return faker.TimeInRange(begin, end).UTC()
+}
+
+// GenRandString generate a random string from the letters set
+// with length n.
+// letters the set of letters to use.
+// n the length of the string.
+// Return a random string.
+func GenRandString(letters []rune, n int) string {
+	s := make([]rune, n)
+	for i := range s {
+		s[i] = letters[GenRandNum(0, int64(len(letters)-1))]
+	}
+	return string(s)
+}
+
+// GenRandDomainLabel generate a random label according to RFC1035
+// Return the string representing a valid domain label.
+func GenRandDomainLabel() string {
+	// See: https://www.rfc-editor.org/rfc/rfc1035
+	//
+	// <domain> ::= <subdomain> | " "
+	// <subdomain> ::= <label> | <subdomain> "." <label>
+	// <label> ::= <letter> [ [ <ldh-str> ] <let-dig> ]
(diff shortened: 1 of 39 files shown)


internal/test/builder/helper/helper.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/google/uuid"
+	"github.com/openlyinc/pointy"
+	"github.com/pioz/faker"
+)
+
+// GenRandNum generate a random bumber >= min and <= max interval.
+// min is the lower boundary interval including the value.
+// max is the higher boundary interval including the value.
+func GenRandNum(min, max int64) int64 {
+	return faker.Int64InRange(min, max+1)
```
avisiedo, line 17: this return interval [min, max) for Int64InRange; that is why is added `+1`.

internal/test/builder/helper/helper.go
```diff
@@ -0,0 +1,296 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/openlyinc/pointy"
+	"github.com/pioz/faker"
+)
+
+// GenRandNum generate a random bumber >= min and <= max interval.
```
avisiedo, line 13:
```suggestion
// GenRandNum generate a random number >= min and <= max interval.
```

internal/test/smoke/suite_test.go
```diff
@@ -0,0 +1,412 @@
...
+		return nil, fmt.Errorf("failure when unmarshalling the information for %s %s", http.MethodPost, url)
+	}
+	return createdDomain, nil
+}
+
+// RunTestCase run test for one specific testcase
+func (s *SuiteBase) RunTestCase(testCase *TestCase) {
+	t := s.T()
+
+	// GIVEN
```
avisiedo, line 201: Does not make sense this comment, because the pre-condition for the tests are provided as the input of the method.

internal/test/smoke/suite_test.go
```diff
@@ -0,0 +1,412 @@
...
+
+	// WHEN
+	resp, err = s.DoRequest(testCase.Given.Method, testCase.Given.URL, testCase.Given.Header, testCase.Given.Body)
+
+	// THEN
+
+	// Check no error
+	require.NoError(t, err)
+	if resp != nil {
+		body, err = io.ReadAll(resp.Body)
```
avisiedo, line 216:
```suggestion
		body, err = io.ReadAll(resp.Body)
		defer resp.Body.Close()
```

internal/test/smoke/suite_test.go
```diff
@@ -0,0 +1,412 @@
...
+
+	// GIVEN
+	var (
+		body []byte
+		resp *http.Response
+		err  error
+	)
+
+	// WHEN
+	resp, err = s.DoRequest(testCase.Given.Method, testCase.Given.URL, testCase.Given.Header, testCase.Given.Body)
```
avisiedo, line 209: It is not checked that for the request only `Body`or `BodyBytes` are used; I suggest to add the check and fail with informative error how to solve for the developer in the future.

internal/test/smoke/suite_test.go
```diff
@@ -0,0 +1,412 @@
...
+		<-exit
+		cancel()
+	}()
+	return ctx, cancel
+}
+
+// TearDownSignalHandler reset the signal handlers
+func TearDownSignalHandler() {
+	signal.Reset(syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+}
```
avisiedo, line 356: Move this block after all the `SuiteBase` method definitions.

aadhikar:
LGTM!

PR #43, opened 2023-11-13
HMS-2437 Use slog as the default logger

cryptomilk:
This introduces slog and sets it up as the default logger. 

If you want a log message, all you have to do is:

```
slog.Info("my info")
slog.Debug("my debug")
...
```

In case you want the logger also in tests, you need to call `logger.InitLogger()`.

log messages look like this in text output, there is also a json output.

```
time=2023-11-13T14:23:14.951+01:00 level=DEBUG msg=routing index=23 route="&{Method:GET Path:/api/idmsvc/v1/signing_keys Name:github.com/podengo-project/idmsvc-backend/internal/api/public.(*ServerInterfaceWrapper).GetSigningKeys-fm}"
```

https://betterstack.com/community/guides/logging/logging-in-go/


tiran:
Could you please include the ticket number in the commit title? Otherwise there is no easy way to get from a commit to a ticket.

IMO all commit titles should start with `HMS-2437 feat:`. After all, the fix commits are part of a feature change.

avisiedo:
I suggest to update `reportError` at `internal/config/config.go` by something like the below:

```golang
func reportError(err error) {
	for _, err := range err.(validator.ValidationErrors) {
		slog.Error(
			"Config validation error",
			slog.String("namespace", err.Namespace()),
			slog.Group("rule",
				slog.String("tag", err.Tag()),
				slog.String("value", fmt.Sprintf("%v", err.Value())),
			),
			slog.String("got", err.Param()),
			slog.String("type", err.Kind().String()),
		)
	}
}
```

internal/infrastructure/datastore/db.go
```diff
@@ -33,18 +33,19 @@ func getUrl(config *config.Config) string {
...
 	if db, err = gorm.Open(pg.Open(dbUrl),
 		&gorm.Config{
-			Logger:                 logger.NewGormLog(cfg),
+			Logger:                 logger.NewGormLog(true),
 			SkipDefaultTransaction: true,
 			// CreateBatchSize:        50,
 			TranslateError: true,
 		}); err != nil {
-		log.Error().Msgf("Error creating database connector: %s", err.Error())
+		slog.Error("Error creating database connector", err.Error())
```
tiran, line 48: This does not work as expected. I'm getting `!BADKEY` with this example code:

```
package main

import (
	"errors"

	"golang.org/x/exp/slog"
)

func main() {
	err := errors.New("Example")
	slog.Error("Error creating database connector", err.Error())
}
```

```
$ bin/testing 
2023/11/14 10:42:55 ERROR Error creating database connector !BADKEY=Example
```

internal/infrastructure/event/consumer.go
```diff
@@ -56,7 +56,7 @@ func NewConsumer(config *config.Kafka) (*kafka.Consumer, error) {
 	if err = consumer.SubscribeTopics(config.Topics, nil); err != nil {
 		return nil, err
 	}
-	log.Info().Msgf("Consumer subscribed to topics: %s", strings.Join(config.Topics, ","))
+	slog.Info("Consumer subscribed to topics", strings.Join(config.Topics, ","))
```
tiran, line 59: `!BADKEY`

internal/infrastructure/event/consumer.go
```diff
@@ -146,10 +149,11 @@ func processConsumedMessage(schemas schema.TopicSchema, msg *kafka.Message, hand
 	if internalTopic == "" {
 		return fmt.Errorf("Topic mapping not found for: %s", *msg.TopicPartition.Topic)
 	}
-	log.Info().
-		Str("Topic name", *msg.TopicPartition.Topic).
-		Str("Requested topic name", internalTopic).
-		Msg("Topic mapping")
+	slog.Info(
+		"Topic mapping",
+		slog.String("Topic name", *msg.TopicPartition.Topic),
```
tiran, line 154: Spaces in struct logging keys create awkward messages like `Topic mapping "Topic name"=topic ...`. I suggest that we use keys that do no need quotes.

internal/infrastructure/event/factory.go
```diff
@@ -22,13 +22,13 @@ func Start(ctx context.Context, config *config.Kafka, handler Eventable) {
 	)
 
 	if consumer, err = NewConsumer(config); err != nil {
-		log.Panic().Msgf("error creating consumer: %s", err.Error())
+		slog.Error("error creating consumer", err.Error())
```
tiran, line 25: `!BADKEY`

avisiedo:
I suggest to extract a lambda at `configCommonMiddleware` at something like the below:

```golang
func logValuesFunc(c echo.Context, v middleware.RequestLoggerValues) error {
	var logLevel slog.Level
	var logAttr []slog.Attr = make([]slog.Attr, 5)
	logAttr = append(logAttr,
		slog.String("request_id", v.RequestID),
		slog.String("method", v.Method),
		slog.String("uri", v.URI),
		slog.Int("status", v.Status),
	)
	if v.Error == nil {
		logLevel = slog.LevelInfo
	} else {
		logLevel = slog.LevelError
		logAttr = append(logAttr, slog.String("err", v.Error.Error()))
	}
	slog.LogAttrs(
		context.Background(),
		logLevel,
		"http_request",
		logAttr...,
	)
	return nil
}
```
Changes on the above:
- Extract the code in a function that can be unit tested isolated.
- The above change print any no error request as informative instead of debug.
- Refactor a little to reduce duplication.

internal/infrastructure/logger/init.go
```diff
@@ -1,18 +1,94 @@
...
+	} else if lvl == "DEBUG" {
+		globalLevel.Set(LevelDebug)
+	} else if lvl == "INFO" {
+		globalLevel.Set(LevelInfo)
+	} else if lvl == "NOTICE" {
+		globalLevel.Set(LevelNotice)
+	} else if lvl == "WARN" {
+		globalLevel.Set(LevelWarn)
+	} else if lvl == "ERROR" {
+		globalLevel.Set(LevelError)
```
tiran, line 92: Contrary to C, you can use a switch statement for string pattern matching. Easier to read.

avisiedo:
I have observed the request id is printed out as an empty string;

```raw
$ ./test/scripts/local-domains-token.sh 
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
X-Frame-Options: SAMEORIGIN
X-Rh-Insights-Request-Id: test_dec4f20e7cfa4e7f940f
Date: Tue, 14 Nov 2023 09:53:47 GMT
Content-Length: 192

{
  "domain_id": "12b4a377-6432-5552-93c8-2ae5836cb1bc",
  "domain_token": "F5d7L2FAhP4._OrBRYLnhLyuK1uYHDL9cfSP5FZmjFlRRqULmNqf1uI",
  "domain_type": "rhel-idm",
  "expiration": 1699962827
}
```

I got the output

```json
{"time":"2023-11-14T10:53:47.780999184+01:00","level":"INFO","msg":"http_request","request_id":"","method":"POST","uri":"/api/idmsvc/v1/domains/token","status":200}
```

I was expecting:

```json
{"time":"2023-11-14T10:53:47.780999184+01:00","level":"INFO","msg":"http_request","request_id":"test_dec4f20e7cfa4e7f940f","method":"POST","uri":"/api/idmsvc/v1/domains/token","status":200}
```

tiran (changes requested):

cryptomilk:
> I suggest to update `reportError` at `internal/config/config.go` by something like the below:
> 
> ```go
> func reportError(err error) {
> 	for _, err := range err.(validator.ValidationErrors) {
> 		slog.Error(
> 			"Config validation error",
> 			slog.String("namespace", err.Namespace()),
> 			slog.Group("rule",
> 				slog.String("tag", err.Tag()),
> 				slog.String("value", fmt.Sprintf("%v", err.Value())),
> 			),
> 			slog.String("got", err.Param()),
> 			slog.String("type", err.Kind().String()),
> 		)
> 	}
> }
> ```

We have a chicken-egg problem here. slog is set up after reading the config, as it relies on the config values. So when we use slog here, it will just log in text form to stdout. Currently we log to stderr ...

avisiedo:
https://github.com/podengo-project/idmsvc-backend/pull/43/#issuecomment-1809888270 , thanks! I didn't realize!

I try to avoid the `init` functions, but what @tiran suggest would work! and I would add to use `func init()` in `main.c` so it would reduce side effects.

tiran:
That would be confusing. `init` functions for logging goes into `internal/infrastructure/logger/` package. `init` functions are fine as long as they do not depend on execution order of other `init` functions.

internal/infrastructure/datastore/db.go
cryptomilk: Fixed

internal/infrastructure/event/consumer.go
cryptomilk: Fixed.

internal/infrastructure/event/consumer.go
cryptomilk: Fixed.

internal/infrastructure/event/factory.go
cryptomilk: Fixed

internal/infrastructure/logger/init.go
cryptomilk: Changed to swtich.

cryptomilk:
I think using `init()` to setup a text logger for stderr is totally fine. I used that now.

cryptomilk:
> I have observed the request id is printed out as an empty string;
> 
> ```
> $ ./test/scripts/local-domains-token.sh 
> HTTP/1.1 200 OK
> Content-Type: application/json; charset=UTF-8
> X-Content-Type-Options: nosniff
> X-Frame-Options: SAMEORIGIN
> X-Rh-Insights-Request-Id: test_dec4f20e7cfa4e7f940f
> X-Xss-Protection: 1; mode=block
> Date: Tue, 14 Nov 2023 09:53:47 GMT
> Content-Length: 192
> 
> {
>   "domain_id": "12b4a377-6432-5552-93c8-2ae5836cb1bc",
>   "domain_token": "F5d7L2FAhP4._OrBRYLnhLyuK1uYHDL9cfSP5FZmjFlRRqULmNqf1uI",
>   "domain_type": "rhel-idm",
>   "expiration": 1699962827
> }
> ```
> 
> I got the output
> 
> ```json
> {"time":"2023-11-14T10:53:47.780999184+01:00","level":"INFO","msg":"http_request","request_id":"","method":"POST","uri":"/api/idmsvc/v1/domains/token","status":200}
> ```
> 
> I was expecting:
> 
> ```json
> {"time":"2023-11-14T10:53:47.780999184+01:00","level":"INFO","msg":"http_request","request_id":"test_dec4f20e7cfa4e7f940f","method":"POST","uri":"/api/idmsvc/v1/domains/token","status":200}
> ```

It is logging `echo.HeaderXRequestID`. This coressponds to: 

```
HeaderXRequestID          = "X-Request-Id"
```

However, we have `X-Rh-Insights-Request-Id: test_094f741b93034bd980b4`. That's why it is empty. However I could implement that ...

docs/dev/TESTING.md
avisiedo, line 1: The change is good! verifying the updated dev doc I have discovered a bug in existing unit tests; jira HMS-3070 ticket created for it.

avisiedo (commented):
Nice change! from my side, all my concerns are resolved. I let @tiran to comment about its concern and suggestions.

During the review a bug was detected in a unit test (repository layer unit tests :cry: ) to be addressed in a different change; HMS-3070 created for it.

internal/config/config.go
```diff
@@ -232,7 +232,7 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 	if clowderConfig.Database != nil && clowderConfig.Database.RdsCa != nil {
 		var err error
 		if rdsCertPath, err = clowderConfig.RdsCa(); err != nil {
-			log.Warn().Err(err).Msg("Cannot read RDS CA cert")
+			slog.Warn("Cannot read RDS CA cert", slog.Any("error", err))
```
tiran, line 235: slog also takes pairs of string / any. `slog.Any("error", err)` and `"error", err` are analogous.

```suggestion
			slog.Warn("Cannot read RDS CA cert", "error", err)
```

The syntax is shorter. I wouldn't use it for complex logging calls. I'd prefer it for logger calls with a single error argument. What do you think, @frasertweedale and @avisiedo ?

internal/infrastructure/service/impl/metrics.go
```diff
@@ -79,19 +88,22 @@ func (srv *metricsService) Start() error {
 	go func() {
 		defer srv.waitGroup.Done()
 		srvAddress := fmt.Sprintf(":%d", srv.config.Metrics.Port)
-		log.Debug().Msgf("metrics: srvAddress=%s", srvAddress)
+		slog.Debug("metrics", slog.String("srvAddress", srvAddress))
 		if err := srv.echo.Start(srvAddress); err != nil && err != http.ErrServerClosed {
-			log.Error().Err(err).Msg("Failed to start metricsService")
+			slog.Error("Failed to start metricsService")
```
tiran, line 93: You are not passing `err` to `slog` here.

internal/config/config.go
```diff
@@ -281,24 +281,26 @@ func Load(cfg *Config) *viper.Viper {
...
-			os.Stderr,
-			"    %s: rule: %v %v, got: '%v' (type: %v)\n",
-			err.Namespace(),
-			err.Tag(), err.Value(), err.Param(), err.Kind(),
+		slog.Error(
+			"Configuration validation error",
+			slog.String("namespace", err.Namespace()),
+			slog.Group("rule",
+				slog.String("tag", err.Tag()),
+				slog.String("value", fmt.Sprintf("%v", err.Value())),
```
tiran, line 300: Is `slog.Any("value", err.Value())` equivalent to `%v` string formatting?

internal/infrastructure/event/factory.go
tiran: Is it fixed? I still see the wrong call in the PR.

tiran, line 25:
```suggestion
		slog.Error("error creating consumer", slog.Any("error", err))
```

internal/infrastructure/event/factory.go
```diff
@@ -22,13 +22,13 @@ func Start(ctx context.Context, config *config.Kafka, handler Eventable) {
...
 		return
 	}
 
 	defer func() {
 		if err := consumer.Close(); err != nil {
-			log.Error().Msgf("error closing consumer: %s", err.Error())
+			slog.Error("error closing consumer", err.Error())
```
tiran, line 31:
```suggestion
			slog.Error("error closing consumer", slog.Any("error", err))
```

internal/infrastructure/logger/gorm_log.go
```diff
@@ -2,92 +2,73 @@ package logger
...
+	elapsedTime := time.Since(begin)
+
+	if err != nil && (!errors.Is(err, gorm.ErrRecordNotFound) ||
+		!l.ignoreRecordNotFound) {
+		sql, rows := fc()
+
+		l.slogger.Log(
+			ctx,
+			LevelTrace,
+			err.Error(),
```
tiran, line 49: This confused me for a sec. `err.Error()` returns a string, which is then used a the primary message in the logger call.

internal/config/config.go
avisiedo: IMHO I'd rather the syntax `slog.Any(...)`, as if the key/pair are not balanced, it is detected in compile time, and would be consistent with more complex structured log messages.

internal/config/config.go
avisiedo: Yes, it seems equivalent; I have just tried.

For the sake of curiosity, `any` is defined as the below:

```raw
// any is an alias for interface{} and is equivalent to interface{} in all ways.
type any = interface{}
```

internal/infrastructure/service/impl/metrics.go
cryptomilk: I added it.

internal/config/config.go
cryptomilk: Fixed.

internal/infrastructure/event/factory.go
cryptomilk: Fixed, thanks

internal/infrastructure/event/factory.go
cryptomilk: Fixed. Thanks.

internal/infrastructure/logger/gorm_log.go
cryptomilk: Well, what should be put as the message instead?

internal/infrastructure/logger/gorm_log.go
```diff
@@ -2,92 +2,73 @@ package logger
...
-			l.logger.Level(zerolog.WarnLevel)
-		}
-	case logger.Error:
-		{
-			l.logger.Level(zerolog.ErrorLevel)
-		}
-	}
-	return l
+func (l *gormLogger) Info(ctx context.Context, msg string, args ...interface{}) {
+	l.slogger.InfoCtx(ctx, msg, args)
```
tiran, line 23: - In latest version, `*Ctx` variants like `InfoCtx` are deprecated in favor of `InfoContext`, `WarnContext`, etc
- gorm uses format-style log messages. You either need to use `Infof()` function or manually `fmt.Sprintf` the message and arguments.
- slog includes the location of a logger call. You have to jump through extra hoops to make slog report the correct location for wrappers. https://github.com/golang/exp/blob/master/slog/example_wrap_test.go has an example how to implement a log wrapper.

cryptomilk: I've implemented it

avisiedo (approved):
I have reviewed the changes and I have tried the error trace for the database, and it LGTM.

I think further changes into the log system can be addressed in future tickets. The goal of the ticket that is the integration with `slog` is achieved.

@cryptomilk ++  nice change! good work!

internal/infrastructure/datastore/db.go
avisiedo: I have checked the last change and it worked; resolving this comment; but if further change were necessary I encourage to open a new ticket to address it.

internal/config/config.go
avisiedo: The last change include this resolution!

Verified by using a `config.yaml` file comenting the `secret` line and the log was printed out with the new update.

Resolving this message.

internal/infrastructure/event/factory.go
avisiedo: good catch! I have reviewed and the last update include this change! resolving the conversation.

internal/infrastructure/event/factory.go
avisiedo: last state is including this change; resolving this conversation.

internal/infrastructure/event/consumer.go
avisiedo: the current state update the initial conversation; resolving this; and addressing a change a didn't realize in a quick ticket; instead of `"strings"` use `"topics"` for the tag of the log message.

internal/infrastructure/event/consumer.go
avisiedo: the current state include the suggested change; resolving the comment.

internal/infrastructure/logger/init.go
avisiedo: the current state updated the chained if by a switch statement; resolving this conversation.

internal/infrastructure/service/impl/metrics.go
avisiedo: The current state include the change to print the error string; resolving this conversation.

internal/infrastructure/logger/gorm_log.go
avisiedo: The change to print out the location is getting unexpected result; a new ticket is addressing this; I got the below:

```raw
{"time":"2023-11-30T17:25:07.731265779+01:00","level":"ERROR","msg":"failed to initialize database, got error failed to connect to `host=localhost user=idmsvc database=idmsvc`: server error (FATAL: password authentication failed for user \"idmsvc\" (SQLSTATE 28P01))"}
```

avisiedo: HMS-3196 filed

avisiedo:
I have added HMS-3196 with additional things; @frasertweedale @tiran feel free to add further changes to be addressed on the ticket.

cryptomilk:
Ping @frasertweedale

avisiedo:
Hi mates! I don't see any further comment in the last weeks so I understand no further concerns about the change; if I were wrong please add anything to HMS-3196; merging the changes

PR #45, opened 2023-11-20
HMS-3070 fix: use EqualError on unit tests

avisiedo:
It was observed during a code review that some unit tests did not raised error on expectations changes. This change fix that situation by using EqualError to check the expectations which checks that an error happens and the string representation match with the string provided at the unit test expectation.


frasertweedale (approved):

frasertweedale:
LGTM.  Thanks!

PR #47, opened 2023-11-21
refactor: Use Golang 1.20 and UBI 9.3

tiran:
- Update from Golang 1.19 to 1.20
- Use latest RHEL 9.3 UBI image
- Update all dependencies. Some updates were blocked by Golang 1.19.


tiran:
github.com/atombender/go-jsonschema@v0.12.1 fails to build with Golang 1.20 because it implements the wrong interface for stable sort. v0.13.0 has a fix, but it requires Golang 1.21.

tiran:
It turns out that go-jsonschema@v.0.13.1 does not actually require Golang 1.21. It works fine with 1.20. I hacked the update into our go.mod and go.sum file and everything just works fine.

cryptomilk (approved):
Looks fine for me. Thanks.

PR #50, opened 2023-11-27
HMS-2678 build: deploy frontend

avisiedo:
This change deploy the frontend together with out custom backend state, so now when we run:

```raw
make ephemeral-deploy
```
we will get the frontend deployed together from the last build image.


tiran:
frontend deployment works after I copied `cp configs/bonfire.example.yaml configs/bonfire.yaml`.

PR #62, opened 2023-11-30
HMS-3155 test: update domain (user)

avisiedo:
This smoke test add an update operation by a user for an existing domain.

It depends on #60

### DIFF
--- a/internal/infrastructure/middleware/validate_api.go
+++ b/internal/infrastructure/middleware/validate_api.go
@@ -36,7 +36,7 @@ func (x xrhiAlwaysTrue) ValidateXRhIdentity(xrhi *identity.XRHID) error {
 func InitOpenAPIFormats() {
 	// TODO Review all the regular expressions
 	openapi3.DefineStringFormat("url", `^https?:\/\/.*$`)
-	openapi3.DefineStringFormat("realm", `^(([A-Z0-9][A-Z0-9\-]*[A-Z0-9])|[A-Z0-9]+\.)*([A-Z]+|xn\-\-[A-Z0-9]+)\.?$`)
+	openapi3.DefineStringFormat("realm", `^([A-Z]|[A-Z][A-Z0-9\-]*[A-Z0-9])(\.([A-Z]|[A-Z][A-Z0-9\-]*[A-Z0-9]))+$`)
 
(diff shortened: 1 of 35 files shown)


internal/infrastructure/middleware/validate_api.go
```diff
@@ -36,7 +36,7 @@ func (x xrhiAlwaysTrue) ValidateXRhIdentity(xrhi *identity.XRHID) error {
 func InitOpenAPIFormats() {
 	// TODO Review all the regular expressions
 	openapi3.DefineStringFormat("url", `^https?:\/\/.*$`)
-	openapi3.DefineStringFormat("realm", `^(([A-Z0-9][A-Z0-9\-]*[A-Z0-9])|[A-Z0-9]+\.)*([A-Z]+|xn\-\-[A-Z0-9]+)\.?$`)
+	openapi3.DefineStringFormat("realm", `^([A-Z]|[A-Z][A-Z0-9\-]*[A-Z0-9])(\.([A-Z]|[A-Z][A-Z0-9\-]*[A-Z0-9]))+$`)
```
frasertweedale, line 39: @tiran can you please eyeball this?  Hostnames with leading digits are permitted by https://datatracker.ietf.org/doc/html/rfc1123#section-2 (e.g. https://9292.nl/) and I think we should not prohibit it in our API.

frasertweedale (commented):
One nit related to the hostname validation regex.

internal/infrastructure/middleware/validate_api.go
frasertweedale: Ah nevermind, this is for the REALM name.  I should read more closely.

frasertweedale:
```
Realm           ::= KerberosString
KerberosString  ::= GeneralString (IA5String)
```

So... I think we should either allow the leading digits in the hostname, or else relax this check closer to the RFC spec, i.e. that it is IA5String (printable ASCII, essentially).  This is notwithstanding the *convention* that realm corresponds to an upper-cased DNS name.  @tiran @cryptomilk what do you think?

avisiedo: Follow up on HMS-3429

avisiedo:

PR #63, opened 2023-11-30
HMS-3157 test: delete domain

avisiedo:
This change add a smoke test for checking the delete domain operation of the API.

Depends on: #60


avisiedo:

PR #64, opened 2023-12-01
infra: Add make container-clean

tiran:
Our container images now have a label `idmsvc-backend` with value `builder` (for temporary Go builder container) or `backend` (for the actual backend container). The builder container is automatically pruned at the end of `make container-build`.

A new make target `container-clean` prunes and removes all local containers with a `idmsvc-backend` label.


avisiedo (approved):
LGTM

PR #68, opened 2023-12-05
feat: IPA client install with mkhomedir and subid

tiran:
Add `--mkhomedir` and `--subid` options to `ipa-client-install` for demo. This configures mkhomedir oddjob and SSSD's subordinate ID provider.


frasertweedale (approved):
obvious ack

PR #71, opened 2023-12-13
HMS-3156 test: update domain (agent)

avisiedo:
This change add the builder for updating a domain by an agent and a smoke test
where the success scenario works correctly.

Depends on: #60


avisiedo:

PR #81, opened 2024-01-10
HMS-3190 feat: Show VCS information on app startup

tiran:
The service now shows version, git commit revision / time stamp, and dirty flag on startup. The information is provided by `runtime/debug` package since Golang 1.18.

`go build` bakes the VCS information into the binary if-and-only-if the main module is references by path to a directory.

### DIFF
--- a/cmd/service/main.go
+++ b/cmd/service/main.go
@@ -30,6 +30,7 @@ func startSignalHandler(c context.Context) (context.Context, context.CancelFunc)
 
 func main() {
 	wg := &sync.WaitGroup{}
+	logger.LogBuildInfo("idmscv-backend")
 	cfg := config.Get()
 	logger.InitLogger(cfg)
 	db := datastore.NewDB(cfg)
--- a/internal/infrastructure/logger/init.go
+++ b/internal/infrastructure/logger/init.go
@@ -2,6 +2,7 @@ package logger
 
 import (
 	"os"
+	"runtime/debug"
 	"strings"
 
 	"github.com/podengo-project/idmsvc-backend/internal/config"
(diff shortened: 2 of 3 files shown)


tiran:
Example output:
```
time=2024-01-10T18:22:33.474+01:00 level=INFO msg=idmsvc-backend Version=(devel) Commit=6cae81ae6565e1185be8a5ee379cc869ece244ff CommitTime=2024-01-10T13:00:39Z Dirty=false
```

cmd/service/main.go
```diff
@@ -28,8 +31,47 @@ func startSignalHandler(c context.Context) (context.Context, context.CancelFunc)
...
+		slog.String("Version", version),
+		slog.String("Commit", revision),
+		slog.String("CommitTime", commitTime),
+		slog.Bool("Dirty", dirty),
+	)
+}
+
 func main() {
 	wg := &sync.WaitGroup{}
+	logApplicationInfo()
```
avisiedo, line 74: I suggest to call this after `logger.InitLogger`; rationalization for getting this message into kibana once we are integrated with clowdWatch.

avisiedo (changes requested):
Overall nice change! I have learned about retrieve the build info.

Only a few suggestions to keep the log functions in the same place, and let reuse the function in additional binaries.

cmd/service/main.go
tiran: I want to log the info before loading configuration, because `config.Get` can fail with a panic. Isn't Kabana using the standard stream output?

avisiedo: my understanding is that it is collected by clowdwatch; envision and content-sources has integration with clowdwatch; asking quickly to be 100% sure.

avisiedo: ok! confirmed it is necessary clowdwatch to see the logs in kibana.

avisiedo: BTW I didn't realize about the panic, thanks for the heads up!

avisiedo: After reviewing the code for the `config.Get`, resolving this conversation; at least we have an entry when the process start in the openshift log if something fails on that operation.

avisiedo (approved):

frasertweedale:
This is very nice.  Thanks @tiran for the code and @avisiedo for the review.

PR #82, opened 2024-01-12
HMS-3196: Some logging improvements

cryptomilk:
This adds some improvements for logging like wiring a config option to print source code locations.


avisiedo (approved):
LGTM

Thanks for this change; I have learned about the AddSource property for the slog handler.

PR #83, opened 2024-01-12
HMS-3138 fix: return 201 when a domain is registered

avisiedo:
Update the returned code by the RegisterDomain handler for a success domain registration, so it return a 201 Created, instead of a 200 Ok status code.


frasertweedale (approved):
obvious ack.  merging.

PR #88, opened 2024-01-17
HMS-3414 build: update ephemeral cluster

avisiedo:
Update to use the new ephemeral cluster and to quickly change to even another cluster if it were necessary just providing the CLUSTER variable from the command line to the make invokation or updating our makefile customizations.

### DIFF
--- a/scripts/mk/ephemeral.mk
+++ b/scripts/mk/ephemeral.mk
@@ -7,6 +7,8 @@ ifeq (,$(APP_NAME))
 $(error APP_NAME is empty; did you miss to set APP_NAME=my-app at your scripts/mk/variables.mk)
 endif
 
+CLUSTER ?= crc-eph.r9lp.p1.openshiftapps.com
+
 APP_COMPONENT ?= backend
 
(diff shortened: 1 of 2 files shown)


scripts/mk/ephemeral.mk
```diff
@@ -79,7 +81,8 @@ ephemeral-login: .old-ephemeral-login ## Help in login to the ephemeral cluster
 
 .PHONY: .old-ephemeral-login
 .old-ephemeral-login:
-	xdg-open "https://oauth-openshift.apps.c-rh-c-eph.8p0c.p1.openshiftapps.com/oauth/token/request"
+	@if [ "$(CLUSTER)" == "" ]; then echo "error:CLUSTER is empty; checkout your secret/private.mk file to setup this value correctly"; exit 1; fi
```
tiran, line 84: IMO this check is not necessary. Line 10 sets a default value. The check only fails when `private.mk` has an empty definition, which should never happen.

avisiedo: thanks mate! updated!

tiran (approved):

PR #89, opened 2024-01-22
HMS-3430 fix: add rbac dependency

avisiedo:
This change introduce the dependency of the idmsvc-backend on the "rbac" service, so it is made available when iqe is launched and when developing the integration with rbac service.


aadhikar (approved):
LGTM!

PR #93, opened 2024-01-23
refactor(HMS-2191): /metrics endpoint

avisiedo:
The reason of this refactor is that the endpoint `/metrics` and port used for it, is dynamically assigned from the environment configuration; this avoid that in the future if such configuration change, the service will not be affected (further than just redeploy a new replicaset).

All the intents to make works with the openapi were not success; the closer was by creating and endpoint `""` for the metrics openapi, and disabling the ruleset to pass the validation, which is not a good approach, using a group with the path for the metrics; given the definition in terms of the openapi is only one endpoint and method, it is extracted directly on the code.

Now the path and port for the metrics are matching with the clowder configuration, and change with it.


.gitignore
```diff
@@ -14,6 +14,7 @@ configs/**
 !configs/config.example.yaml
 !configs/config.ci.yaml
 !configs/bonfire.example.yaml
+!configs/cdappconfig.json
```
tiran, line 17: Why do you ignore the file and add the file to git?

configs/cdappconfig.json
```diff
@@ -0,0 +1,63 @@
+{
+  "database": {
+    "adminPassword": "postgres",
+    "adminUsername": "postgres",
+    "hostname": "localhost",
+    "name": "idmsvc",
+    "password": "idmsvc",
+    "port": 5432,
+    "sslMode": "disable",
+    "username": "idmsvc"
```
tiran, line 10: Could you please use different values for name, user, and password like in `deployments/docker-compose.yaml` ?

configs/cdappconfig.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "kafka": {
+    "brokers": [
+      {
+        "hostname": "localhost",
+        "port": 9092
+      }
+    ],
+    "topics": []
+  },
```
tiran, line 38: On stage, the clowder config does not have a Kafka entry. See #97

tiran (changes requested):
Now that you are removing the auto-generated code, can we also remove `metrics.openapi.yaml` from idmsvc-api?

.gitignore
avisiedo: This configuration file is an example for the format that we receive from clowder. It includes the port and path for the metrics endpoint.
The application use this to dynamically set the endpoint, so if it is changed in the future for whatever is the reason, our application automatically change this too.

configs/cdappconfig.json
avisiedo: Thanks, updated!

configs/cdappconfig.json
avisiedo: Thanks! updated

avisiedo:
About the changes on the openapi spec, I think you refer to this change: https://github.com/podengo-project/idmsvc-api/pull/15/files

tiran (approved):

PR #95, opened 2024-01-23
HMS-2182 refactor: enforce identity order

avisiedo:
This change provide a deterministic order when the enforce identity middleware is applied, when more than one predicate is provided.


tiran (approved):
LGTM

Golang does not retain insertion order of maps. Arrays are sorted.

PR #97, opened 2024-01-26
HMS-3474 fix: Check for clowder AppConfig.Kafka for nil

tiran:
Clowder configuration may have AppConfig.Kafka set to `nil`. The function `NewTopicTranslationWithClowder` was not checking the value for nil and crashed when it attempted to iterate over cfg.Kafka.Topics.


avisiedo:
LGTM

avisiedo (approved):
LGTM

Thanks for the change

PR #99, opened 2024-01-26
HMS-3190 fix: Include VCS info in container binaries

tiran:
Do not exclude `.git` directory from Docker COPY, so `go build` is able to figure out and include VCS information in binaries that are built inside a container.

### DIFF
--- a/build/package/Dockerfile
+++ b/build/package/Dockerfile
@@ -8,6 +8,7 @@ ENV OPENSSL_FORCE_FIPS_MODE=1
 WORKDIR /go/src/app
 COPY . .
 USER 0
+RUN git log -1
 RUN make get-deps build
 
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,6 +14,5 @@ configs
 # File generated when running unit tests
 coverage.out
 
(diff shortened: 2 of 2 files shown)


build/package/Dockerfile
```diff
@@ -8,6 +8,7 @@ ENV OPENSSL_FORCE_FIPS_MODE=1
 WORKDIR /go/src/app
 COPY . .
 USER 0
+RUN git log -1
```
tiran, line 11: The additional command logs the latest revision and ensures that the container has access to VCS information.

avisiedo (approved):
LGTM

Checked on dev cluster:

```raw
Defaulted container "idmsvc-backend-service" out of: idmsvc-backend-service, crcauth, db-migrate-init (init)
time=2024-01-26T09:20:41.048Z level=INFO msg=idmscv-backend Version=(devel) Commit=4f35882eab30a60bd90824f5c54e235af7c020a2 CommitTime=2024-01-26T08:30:16Z Dirty=true
⇨ http server started on [::]:9000
⇨ http server started on [::]:8000
```

PR #100, opened 2024-01-26
infra: Use minimal UBI image with updates

tiran:
The `ubi9-minimal` image contains everything we need to run our app and enough infrastructure for interactive debugging. Also install security updates with `microdnf`.


avisiedo (approved):
LGTM

Clair report in quay.io are green! thanks for the change.

PR #110, opened 2024-02-02
HMS-3461 fix: Add service to Clowder template

tiran:
Add service to route HTTP and metrics.


tiran:
After some internal discussions and investigation, it looks like we do not need a service definition after all. I'm keeping the PR around until I have confirmation from the gateway team.

frasertweedale:
That seems to be correct.

Whereas content-sources ClowdApp definition does have a `Service`:

https://github.com/content-services/content-sources-backend/blob/main/deployments/deployment.yaml#L530-L550

... the launch service does not:  https://github.com/RHEnVision/provisioning-backend/blob/main/deploy/clowdapp.yaml

tiran:
The change is not needed.

PR #116, opened 2024-02-07
fix(HMS-3559): fix dependencies

avisiedo:
- Provides deterministic behavior when using go1.20 and g1.21
- Downgrade some tools that requires go1.21


avisiedo:
Closing this; follow up on https://github.com/podengo-project/idmsvc-backend/pull/117

PR #117, opened 2024-02-07
HMS-3559 infra: Fix Go version confusion / pr-check

tiran:
`tools/go.mod` was using Go version 1.21, but our build container is limited to 1.20. Address build confusion by downgrading to a working go-jsonschema version and pinning `go mod tidy` to 1.20.

Introduce `pr-check` to run most important and fast linting, formatting, and testing steps.

### DIFF
--- a/scripts/mk/go-rules.mk
+++ b/scripts/mk/go-rules.mk
@@ -60,8 +60,8 @@ run: $(BIN)/service .compose-wait-db ## Run the api & kafka consumer locally
 # See: https://go.dev/doc/modules/managing-dependencies#synchronizing
 .PHONY: tidy
 tidy:  ## Synchronize your code's dependencies
-	go mod tidy
-	cd tools && go mod tidy
+	go mod tidy -go=$(GO_VERSION)
+	cd tools && go mod tidy -go=$(GO_VERSION)
 
 .PHONY: get-deps
 get-deps: ## Download golang dependencies
@@ -119,6 +119,12 @@ $(patsubst cmd/%,$(BIN)/%,$(wildcard cmd/*)): $(shell find $(PROJECT_DIR)/cmd -t
 # $(shell find "$(EVENT_MESSAGE_DIR)" -type f -name '*.go'): $(SCHEMA_YAML_FILES)
 # 	$(MAKE) gen-event-messages
 
+.PHONY: pr-check
+pr-check: ## Run common checks before submitting a PR
+	$(MAKE) go-fmt
+	$(MAKE) tidy vet
+	$(MAKE) generate-api generate-mock install-go-tools
+	$(MAKE) build test-unit
 
 ############### TOOLS
 
--- a/scripts/mk/variables.mk
+++ b/scripts/mk/variables.mk
@@ -16,6 +16,8 @@ COMPOSE_FILE ?= $(PROJECT_DIR)/deployments/docker-compose.yaml
 
 CONTAINER_IMAGE_BASE ?= quay.io/$(firstword $(subst +, ,$(QUAY_USER)))/$(APP_NAME)-$(APP_COMPONENT)
 
+GO_VERSION ?= 1.20
+
 # Tools and their dependencies
 # Build dependencies
 TOOLS_BIN := tools/bin
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,9 +12,3 @@ repos:
     rev: v0.5.1
     hooks:
       - id: golangci-lint
-  # TODO This linter is not set up correctly
-  - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.21.0
-    hooks:
-      - id: check-metaschema
-        files: ^api/event/.*\.yaml$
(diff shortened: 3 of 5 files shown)


tiran:
This PR is basically the same as #116 but with additional bells and whistles.

scripts/mk/go-rules.mk
```diff
@@ -60,8 +60,18 @@ run: $(BIN)/service .compose-wait-db ## Run the api & kafka consumer locally
...
+
+.PHONY: container-go-tidy
+container-tidy:
+	$(CONTAINER_ENGINE) run \
+	  --rm \
+	  --userns=keep-id:uid=1001,gid=1001 \
+	  -v $(PROJECT_DIR):/opt/app-root/src/idmsvc-backend:Z \
+	  -w /opt/app-root/src/idmsvc-backend \
+	  $(GO_TOOLSET_CONTAINER) \
+	  make tidy
```
avisiedo, line 74: I'd rather to avoid leverage of container on this file, and I think the change in `tidy` rule would be enough.

scripts/mk/variables.mk
```diff
@@ -16,6 +16,9 @@ COMPOSE_FILE ?= $(PROJECT_DIR)/deployments/docker-compose.yaml
 
 CONTAINER_IMAGE_BASE ?= quay.io/$(firstword $(subst +, ,$(QUAY_USER)))/$(APP_NAME)-$(APP_COMPONENT)
 
+GO_VERSION ?= 1.20
```
avisiedo, line 19: I like this to reduce the burden when upgrading the golang version; thanks! :+1:

scripts/mk/go-rules.mk
```diff
@@ -119,6 +129,12 @@ $(patsubst cmd/%,$(BIN)/%,$(wildcard cmd/*)): $(shell find $(PROJECT_DIR)/cmd -t
 # $(shell find "$(EVENT_MESSAGE_DIR)" -type f -name '*.go'): $(SCHEMA_YAML_FILES)
 # 	$(MAKE) gen-event-messages
 
+.PHONY: pr-check
+pr-check: ## Run common checks before submitting a PR
+	$(MAKE) go-fmt
+	$(MAKE) container-tidy vet
+	$(MAKE) generate-api generate-mock install-go-tools
+	$(MAKE) build test-unit
```
avisiedo, line 137: I like the idea of `pr-check` rule; I suggest the below:

```suggestion
pr-check: ## Run common checks before submitting a PR
	[ ! -e "$(PROJECT_DIR)/tools/bin" ] || rm -rvf "$(PROJECT_DIR)/tools/bin"
	$(MAKE) -j1 compose-clean clean install-tools build lint compose-up test container-build
```

* vet is actually invoked by `make lint` (it is enabled on `.golangci.yaml` file). Indeed `make vet` rule could be removed.
* for the formatting it is setup `gofumpt` see: `.vscode/settings.json` and `docs/CONTRIBUTING.md` files.

scripts/mk/variables.mk
avisiedo: I suggest to add a note at `go.mod` in the base repository to address where to update when upgrading the golang version for the future.

avisiedo (changes requested):
thanks for the changes; only a few request changes.

avisiedo:
Please reword the commit to fit docs/CONTRIBUTING.md

It would be `infra(HMS-3559): Fix Go version confusion / pr-check`

scripts/mk/variables.mk
tiran: `go.mod` is auto-generated and will be overwritten

scripts/mk/go-rules.mk
tiran: - `make lint` uses pre-commit. I don't particularly like the tool, because it is too magic to debug and yet another layer of indirection.
- I prefer `pr-check` to be lightweight and not involve compose and container stuff.

tiran:
> Please reword the commit to fit docs/CONTRIBUTING.md
> 
> It would be `infra(HMS-3559): Fix Go version confusion / pr-check`

I thought that we agreed on style like `HMS-3559 infra: ...`, so the ticket number appears in the beginning.

frasertweedale:
FWIW, I do not care about commit messages conforming to some schema, as long as the required info is present.  We don't have any tooling that depends on it, AFAIK.

scripts/mk/go-rules.mk
avisiedo: I don't full agree; I think in freeipa-operator repository I made the mistake of adding many linters and validations (no leveraging pre-commit) on the repository and pipeline because I didn't know the potential of pre-commit.

After that learning, I have tried to delegate as much as possible on the pre-commit configuration (using or not as a hook for git-commit, which I leave to the decision of each developer); the reason is because it is pretty comfortable to add new linters from existing hooks and no reinvent the wheel; everything is installed by `make install-tools` as indicated on the getting started doc (once task).

I agree on extract `compose-clean` and `compose-up` from the proposed rule as that would make it incompatible with the execution of the `make pr-check` from the github pipeline, delegating on the developer to be aware that it should be started the local infrastructure before run the rule (be aware `make test` now execute `test-unit` and `test-smoke` rules for the matching type of tests, and `test-smoke` require the local infrastructure).

Said that, accepting the changes and it can always be changed in the future if it were necessary.

avisiedo (approved):

PR #119, opened 2024-02-08
HMS-3481 fix: Use vault for app-secret

tiran:
app-interface and vault now provide secret for ephemeral environment, too. Use the pre-defined secret by default.


avisiedo (approved):
LGTM

- [x] Deploy by `bonfire deploy idmsvc` worked with no additional steps.
- [x] Removing the secret at `secrets/ephemeral/*`, deploy by `make ephemeral-build-deploy` worked with no additional steps.

PR #121, opened 2024-02-12
HMS-2694 refactor: Split token package

tiran:
Split the token package into three separate packages and move the HKDF code into a new `secrets` package.

- `hostconf_token` deals with host configuration token
- `hostconf_jwk` handles private and public JWKs for host configuration token
- `domain_token` deals with domain registration tokens
- new `secrets` expands and extracts secrets from the main app secret
- secrets are now part of `config` structure in an untagged struct member (therefore not handled by viper)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .DS_Store
 .cicd_bootstrap.sh
-secrets
+/secrets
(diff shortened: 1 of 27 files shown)


frasertweedale:
LGTM, merged.

PR #126, opened 2024-02-12
feat(HMS-3521): Add configuration to enable/disable RBAC middleware

avisiedo:
This change add the configuration and skeleton for the RBAC middleware and unit tests.

The default value at config will be enabled in the code, but will be disabled until the implementation is full finished at HMS-3528

Checked by:

- `APP_ENABLE_RBAC=false make test-smoke` # run successfully
- `APP_ENABLE_RBAC=true make test-smoke`  # Fail with 403
- `make test` # Run successfully

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -140,7 +140,6 @@ linters:
     - testifylint
     - unused
     - whitespace
-    - wrapcheck
     # WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner. Replaced by revive.
     # - golint
     # # - asasalint
--- a/configs/config.ci.yaml
+++ b/configs/config.ci.yaml
@@ -80,3 +80,6 @@ app:
   # main secret for various MAC and encryptions like domain registration
   # token and encrypted private JWKs. "random" generates an ephemeral secret.
   secret: random
+  # Enable/Disable RBAC verification
+  # TODO remove override when HMS-3521 is implemented
+  enable_rbac: false
--- a/configs/config.example.yaml
+++ b/configs/config.example.yaml
@@ -82,3 +82,6 @@ app:
   # main secret for various MAC and encryptions like domain registration
   # token and encrypted private JWKs. "random" generates an ephemeral secret.
   secret: random
+  # Enable/Disable RBAC verification
+  # TODO remove override when HMS-3521 is implemented
+  enable_rbac: false
--- a/deployments/clowder.yaml
+++ b/deployments/clowder.yaml
@@ -215,3 +215,8 @@ parameters:
     value: "false"
     description: |
       A flag to suspend execution of 'db-tool jwk refresh' cron job.
+  # TODO HMS-3528 Set "true" or remove the feature flag
+  - name: APP_ENABLE_RBAC
+    value: "false"
+    description: |
+      It allows to enable / disable RBAC middleware.
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -20,12 +20,15 @@ func TestSetDefaults(t *testing.T) {
 		setDefaults(v)
 	})
 
+	// TODO HMS-3609 There are many more properties to check; complete the test
 	assert.Equal(t, DefaultWebPort, v.Get("web.port"))
 	assert.Equal(t, "info", v.Get("logging.level"))
 	assert.Equal(t, "http://localhost:8010/api/inventory/v1", v.Get("clients.host_inventory_base_url"))
 	assert.Equal(t, DefaultTokenExpirationTimeSeconds, v.Get("app.token_expiration_seconds"))
 	assert.Equal(t, PaginationDefaultLimit, v.Get("app.pagination_default_limit"))
 	assert.Equal(t, PaginationMaxLimit, v.Get("app.pagination_max_limit"))
+
+	assert.Equal(t, DefaultEnableRBAC, v.Get("app.enable_rbac"))
 }
 
 func TestSetClowderConfiguration(t *testing.T) {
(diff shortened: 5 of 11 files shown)


internal/infrastructure/router/public.go
```diff
@@ -93,6 +94,16 @@ func newGroupPublic(e *echo.Group, c RouterConfig) *echo.Group {
...
+	rbacMiddleware := middleware.DefaultNooperation
+	if config.Get().Application.EnableRBAC {
+		rbacMiddleware = middleware.RBACWithConfig(
+			&middleware.RBACConfig{
+				Skipper: nil,
+				// TODO HMS-3522
+				PermissionMap: middleware.RBACMap{},
+			},
+		)
+	}
```
tiran, line 106:
```suggestion
	if config.Get().Application.EnableRBAC {
		rbacMiddleware = middleware.RBACWithConfig(
			&middleware.RBACConfig{
				Skipper: nil,
				// TODO HMS-3522
				PermissionMap: middleware.RBACMap{},
			},
		)
	} else {
		rbacMiddleware := middleware.DefaultNooperation
	}
```

internal/infrastructure/middleware/rbac.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/labstack/echo/v4"
+	echo_middleware "github.com/labstack/echo/v4/middleware"
+)
+
+// TODO HMS-3522 Add RBACMap and implement its methods:
+//
+// Add(route string, method string, permission string) RBACMap
+// Get(route string, method string) string
+type RBACMap map[string]map[string]string
```
tiran, line 14: Please document what the RBACMap contains. Is it `route -> method -> permission` or `method -> route -> permission`?

internal/config/config_test.go
```diff
@@ -20,12 +20,15 @@ func TestSetDefaults(t *testing.T) {
 		setDefaults(v)
 	})
 
+	// TODO There are many more properties to check; complete the test
```
tiran, line 23: I suggest that you turn these `TODO` comments into a refactoring ticket.

tiran (changes requested):
Suggestion to make the PR shorter:
- Set `DefaultEnableRBAC = false` with a comment that RBAC is currently disabled because the feature is under development
- Remove all other places that set `enable_rbac: false`

deployments/clowder.yaml
```diff
@@ -188,3 +188,8 @@ parameters:
     description: |
       It allows to validate API requests by using the
       service OpenAPI specification.
+  # TODO HMS-3528 Set "true" or remove the feature flag
+  - name: APP_ENABLE_RBAC
+    value: "false"
+    description: |
+      It allows to enable / disable RBAC middleware.
```
tiran, line 195: I think this should be set to `true` by default. The app-interface change will set it to `false` for stage and ephemeral. You also have to pass the parameter as env var to the container.

configs/config.ci.yaml
```diff
@@ -80,3 +80,6 @@ app:
   # main secret for various MAC and encryptions like domain registration
   # token and encrypted private JWKs. "random" generates an ephemeral secret.
   secret: random
+  # Enable/Disable RBAC verification
+  # TODO Change to true when the feature is ready
```
tiran, line 84: Let's include the ticket number, so it is easier to find the places that have to be touched later.

```suggestion
  # TODO remove override when HMS-3521 is implemented
```

configs/config.example.yaml
```diff
@@ -82,3 +82,6 @@ app:
   # main secret for various MAC and encryptions like domain registration
   # token and encrypted private JWKs. "random" generates an ephemeral secret.
   secret: random
+  # Enable/Disable RBAC verification
+  # TODO Change to true when the feature is ready
```
tiran, line 86:
```suggestion
  # TODO remove override when HMS-3521 is implemented
```

tiran (changes requested):

.golangci.yaml
```diff
@@ -140,7 +140,7 @@ linters:
     - testifylint
     - unused
     - whitespace
-    - wrapcheck
+    # - wrapcheck
```
frasertweedale, line 143: If we disable it, why not just delete?  Or please put a comment explaining why it is commented :)

deployments/clowder.yaml
frasertweedale: or we just remove the definition (now or later) - absence shall imply "enabled".

frasertweedale (approved):
One very small nit.  Not a show-stopper.  Please consider it, if you decide to change it, that's fine.  If not, that's OK too.  Then you can merge this PR :)

deployments/clowder.yaml
avisiedo: Indeed, I recognize it is confusing; my idea was "enabled by default" when everything is finished, but at the same time, don't brake anything until it is ready to work, so:

- When nothing is specified, the default value in the code (configuration) is enabled.
- In the config.yaml it is set the value as `false`, so it does not impact to the developer in other areas.
- In the makefile it is defined optional to `false`, to make possible to override the value from the CLI `APP_ENABLE_RBAC=true make run`, or the opposite.
- In the `clowder.yaml`, I had to define the parameter to be used from app-interface; it is useful when deploying in ephemeral so nothing is broken in IQE.
- In the app-interface, I force to false (until HMS-3528 is merged that enable everything); probably this is where no change is necessary at all; but when I thought about it at the beginning it made sense.

.golangci.yaml
avisiedo: Honestly I don't remember why I commented the line. I leave it to keep the list of the linters on the file, but the reference before the `enable:` section is enough.

I have removed the line.

deployments/clowder.yaml
avisiedo: my conclusion; if we set this `true`, we will brake the IQE tests until the middleware is implemented (bear in mind they deploy by `bonfire deploy [--frontends true] idmsvc` with no additional parameters (which is right; what is in the makefiles is to make the developers life easier so the bonfire options are automatically filled to get the current commit state deployed); IMHO I would leave this parameter as it is.

Another question is if the app-interface MR is necessary; because we are managing the dafault value in the openshift template until we get the middleware ready. I think it could be cancelled.

What do you think?

avisiedo: Another thought in relation with the app-interface: from a perspective of a gradual deployment (1) change the deployment parameter in ephemeral (or remove), and when the feature looks working correctly, (2) change the deployment parameter (or remove) in stage, and get everything working.

So `app-interface` MR is necessary too.

tiran: The app-interface MR has landed

PR #127, opened 2024-02-12
build(HMS-3187): dependabot dependencies

avisiedo:
Close #126 
Close #123 
Close #122 
Close #111


tiran:
I'm merging the dependabot PRs for x/crypto, validator, and platform-go-middlewares. We downgraded yq on purpose, so the update will be skipped until we have Golang 1.21.

PR #130, opened 2024-02-15
HMS-2694 feat: Add boilerplate for JWK key refresh

tiran:
The `db-tool` command now has a subcommand to refresh JWKs. The subcommand doesn't do anything besides printing a string. The actual refresh code will be implemented later.

- add dummy implementation of `db-tool jwk refresh`
- update cloweder template to run refresh in init container and as periodic job. The cron job runs every hour, which makes testing and debug a bit easier. Later one we could reduce the schedule to daily.
- run `db-tool jwk refresh` during `make db-migrate-up`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -31,6 +31,18 @@
             }
         },
         {
+            "name": "db-tool-jwk-refresh",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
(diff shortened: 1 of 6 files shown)


frasertweedale (approved):

frasertweedale:
Uh, needs rebase?  (dunno why, rebase was clean on my local clone...)

PR #134, opened 2024-02-19
HMS-2694 feat: Store Hostconf JWKs in DB

tiran:
The Hostconf JWKs are now stored and retrieved from database. Keys are
created and refreshed on startup, either by an init container or by
`make db-migrate-up` call.

The `db-tool` has commands to list, purge, refresh, and revoke keys.

```
$ bin/db-tool jwk
Hostconf JWK management

Available Commands:
  list        List JWKs
  purge       Purge expired JWK
  refresh     Refresh or create JWKs
  revoke      Revoke a JWK
```

### DIFF
--- a/internal/usecase/repository/hostconf_jwk_repository.go
+++ b/internal/usecase/repository/hostconf_jwk_repository.go
@@ -1,12 +1,12 @@
 package repository
 
 import (
-	"encoding/json"
 	"fmt"
 	"time"
 
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/podengo-project/idmsvc-backend/internal/config"
+	internal_errors "github.com/podengo-project/idmsvc-backend/internal/errors"
 	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/token/hostconf_jwk"
 	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/token/hostconf_jwk/model"
 	"github.com/podengo-project/idmsvc-backend/internal/interface/repository"
@@ -17,9 +17,6 @@ var notImplementedError = fmt.Errorf("TODO: not implemented")
 
 type hostconfJwkRepository struct {
 	config *config.Config
-	// TODO: store JWKs in database
-	signingKeys []jwk.Key
-	publicKeys  []string
 }
 
 func NewHostconfJwkRepository(cfg *config.Config) repository.HostconfJwkRepository {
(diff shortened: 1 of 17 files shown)


internal/usecase/repository/hostconf_jwk_repository.go
```diff
@@ -17,77 +17,140 @@ var notImplementedError = fmt.Errorf("TODO: not implemented")
...
+	}
+	// find JKW by unique kid
+	if err = db.
+		Where("key_id = ?", kid).
+		First(&hcjwk).
+		Error; err != nil {
+		return nil, err
+	}
+
+	// Set encrypted JWK field to NULL and safe the new object
```
frasertweedale, line 56: nit: typo safe -> save.

Also, it probably justifies a comment to explain that revocation is accomplished simply
by removing the (encrypted) private key. This was (for me) an unexpected solution.  It's
perfectly fine but probably deserves a brief commentary.

internal/usecase/repository/hostconf_jwk_repository.go
```diff
@@ -17,77 +17,140 @@ var notImplementedError = fmt.Errorf("TODO: not implemented")
...
+func (r *hostconfJwkRepository) PurgeExpiredJWKs(db *gorm.DB) (hcjwks []model.HostconfJwk, err error) {
+	if db == nil {
+		return nil, internal_errors.NilArgError("db")
+	}
+
+	// list JWKs, order by id to have determinstic sorting
+	now := time.Now()
+	if err = db.
+		Order("id").
+		Where("expires_at <= ?", now). // use SQL NOW()?
```
frasertweedale, line 90: Should we allow for a small clock skew here?  We should be generating and switching to a new key ~30 days before expiry, but just in case?

internal/infrastructure/datastore/hostconf_jwk_refresh.go
```diff
@@ -0,0 +1,234 @@
...
+	)
+	db = NewDB(r.cfg)
+	defer Close(db)
+
+	if tx = db.Begin(); tx.Error != nil {
+		return tx.Error
+	}
+	defer tx.Rollback()
+
+	r.timestamps()
```
frasertweedale, line 216: We are invoking `timestamps()` here... just to log some configuration (renewal threshold, and validity duration), the current time, and the current time + those values?

Unsure why this is useful in the context of the list command.  I think it would be more useful to have a per-JWK indication that the key is expired, revoked, or due for renewal (similar to what occurs in `refresh` command).

internal/usecase/repository/hostconf_jwk_repository.go
tiran: I have moved the revocation 'logic' to the hostconf JWK model and added comments to the function and model.

I think it is an elegant solution. We don't need the private key of a revoked JWK anyway. The approach does not need an additional column in the DB and we cannot accidentally use the private key of a revoked JWK.

internal/usecase/repository/hostconf_jwk_repository.go
tiran: A small clock skew would make testing more complicated. Besides an hour clock skew doesn't make a difference when the renewal window is about a month.

internal/infrastructure/datastore/hostconf_jwk_refresh.go
tiran: Good point. I'm now included `expiresAfter` and `renewlAfter` information in the key log lines:

```
msg="Hostconf JWK configuration" now=2024-02-21T09:37:47.000Z renewAfter=2024-03-22T09:37:47.000Z renewThreshold=720h0m0s expiresAfter=2024-05-21T09:37:47.000Z validity=2160h0m0s
msg="JWKs in database" length=2
msg="Hostconf JWK" kid=mbDs3TIE publickey=revoked privatekey=revoked expires=2024-05-16T09:02:51.000Z expiresAfter=2024-05-21T09:37:47.000Z renewalAfter=2024-03-22T09:37:47.000Z
msg="Hostconf JWK" kid=PeBB1c46 publickey=valid privatekey=valid expires=2024-05-19T08:43:51.000Z expiresAfter=2024-05-21T09:37:47.000Z renewAfter=2024-03-22T09:37:47.000Z
```

internal/usecase/repository/hostconf_jwk_repository.go
frasertweedale: OK, I'm satisfied.

internal/usecase/repository/hostconf_jwk_repository.go
frasertweedale: Thanks for that.  I agree it's elegant, and the implementation is easier to understand now.

frasertweedale (approved):

PR #137, opened 2024-02-21
refactor: Remove unnecessary BeforeCreate hook

tiran:
There is no need to populate CreatedAt and UpdatedAt fields with a custom hook. Gorm maintains the fields automatically for us.

See: https://gorm.io/docs/models.html#gorm-Model


frasertweedale:
`-35,+0`

You have my ack!

frasertweedale (approved):

PR #138, opened 2024-02-21
HMS-3603 Implement support for Cloudwatch

cryptomilk:
Implement support for Cloudwatch.

ToDo:
- [x] Unit tests for Slog Meta Handler
- [ ] Test with real cloudwatch (can only be done in staging area)


internal/infrastructure/logger/meta_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	return ret
+}
+
+// WithAttrs returns a new TextHandler whose attributes consists
+// of h's attributes followed by attrs.
+func (h *SlogMetaHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	var newMeta = &SlogMetaHandler{handlers: []slog.Handler{}}
+	for i := range h.handlers {
+		h.handlers[i] = h.handlers[i].WithAttrs(attrs)
```
tiran, line 41: Shouldn't this append the new handler to newMeta instead of modifying the current meta handler?

internal/infrastructure/logger/meta_handler_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	w := &stringWriter{buffer: &bytes.Buffer{}}
+	slogMetaHandlerCreate(mh, w)
+
+	logger := slog.New(mh)
+	logger.WithGroup("group")
+	logger.Error("z", slog.String("foo", "bar"))
+
+	assert.Equal(t, "time=NODATE level=ERROR msg=z group.foo=bar\n", w.String())
```
tiran, line 118: This result is wrong. `WithGroup` should not modify the original logger. Instead `logger.WithGroup()` returns a new logger that has grouping enabled.

internal/infrastructure/logger/meta_handler.go
cryptomilk: Yep, fixed it. Thanks!

internal/infrastructure/logger/meta_handler_test.go
cryptomilk: Yep, we fixed that too. It is the same bug as with WithAttrs.

internal/infrastructure/logger/meta_handler.go
tiran: +1

I like your approach with `make()`. It's more efficient than appending each handler.

cryptomilk: Yes, good catch. Fixed :-)

avisiedo (approved):
LGTM

PR #147, opened 2024-03-06
HMS-3603 feat(logging): Add a log message about the logging setup

cryptomilk:
This will always log an info message on the console on startup and tell us how logging will be configured.

Example:
`time=2024-03-06T11:31:24.289+01:00 level=INFO msg=idmsvc-backend Logging.Console=true Logging.Type=null Logging.Level=warn`

### DIFF
--- a/internal/infrastructure/logger/init.go
+++ b/internal/infrastructure/logger/init.go
@@ -1,6 +1,7 @@
 package logger
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"runtime/debug"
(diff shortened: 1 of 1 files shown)


cryptomilk:
Logging.Type should be "cloudwatch" on the staging server.

internal/infrastructure/logger/init.go
```diff
@@ -48,6 +48,16 @@ func InitLogger(cfg *config.Config) {
...
+	// This will be always logged on service startup
+	slog.Info(
+		"idmsvc-backend",
+		slog.Group("Logging",
+			slog.Bool("Console", cfg.Logging.Console),
+			slog.String("Type", cfg.Logging.Type),
+			slog.String("Level", cfg.Logging.Level),
+		),
+	)
+
```
avisiedo, line 60: I guess we would need to move the block at the end of the function, so the metahandler and cloudwatch are set as the default slog, and the level of the log.

avisiedo (changes requested):
only a change to move the block at the end of the function; good idea the trace.

internal/infrastructure/logger/init.go
avisiedo: my bad, I understood in a different way :+1:

cryptomilk:
I've added a log message at the end too and also added Location to the initial log line

avisiedo (approved):
LGTM

PR #148, opened 2024-03-07
HMS-3429: fix realm name check

frasertweedale:
Relax the realm name check to better conform to the Kerberos definition (RFC 4120).  We now require it to consist of printable ASCII characters.  This still goes beyond the type in the spec (IA5String) but is a reasonable compromise between handling unusual but legitimate real world scenarios, and avoiding dealing with data that may be corrupt, malicious or otherwise indicative of a serious error.

Fixes: https://issues.redhat.com/browse/HMS-3429


avisiedo:
sorry for the delay and thanks for your patient.

I will start tomorrow with this review.

tiran (approved):
LGTM

PR #157, opened 2024-03-14
Fix: logging location needs simple template for boolean

tiran:
Fixes:
```
* spec.deployments[0].podSpec.env[2].value: Invalid value: "boolean": spec.deployments[0].podSpec.env[2].value in body must be of type string: "boolean"
```


aadhikar (approved):
LGTM!

PR #162, opened 2024-03-20
fix(HMS-3731): workaround for cloudwatch type field

avisiedo:
This workaround would fix the cloudwatch situation, so
the type field is filled with `"coudwatch"` string when
the cloudwatch values are merged on the service configuration.

When https://github.com/RedHatInsights/clowder/pull/627 PR
is merged, we can revert the first commit of this PR.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -290,6 +290,11 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 		v.Set("logging.cloudwatch.group", clowderConfig.Logging.Cloudwatch.LogGroup)
 		v.Set("logging.cloudwatch.region", clowderConfig.Logging.Cloudwatch.Region)
 		v.Set("logging.cloudwatch.secret", clowderConfig.Logging.Cloudwatch.SecretAccessKey)
+		// TODO Delete the block below when the below PR is merged
+		// See: https://github.com/RedHatInsights/clowder/pull/627
+		if clowderConfig.Logging.Type == "" && len(clowderConfig.Logging.Cloudwatch.SecretAccessKey) > 0 {
+			v.Set("logging.type", "cloudwatch")
+		}
 	}
 
 	// Metrics configuration
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -103,6 +103,7 @@ func TestSetClowderConfiguration(t *testing.T) {
 	assert.Equal(t, "testgroup", v.Get("logging.cloudwatch.group"))
 	assert.Equal(t, "testsecret", v.Get("logging.cloudwatch.secret"))
 	assert.Equal(t, "testaccesskeyid", v.Get("logging.cloudwatch.key"))
+	assert.Equal(t, "cloudwatch", v.Get("logging.type"))
 }
 
 func TestHasKafkaBrokerConfig(t *testing.T) {


internal/config/config.go
```diff
@@ -290,6 +290,11 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 		v.Set("logging.cloudwatch.group", clowderConfig.Logging.Cloudwatch.LogGroup)
 		v.Set("logging.cloudwatch.region", clowderConfig.Logging.Cloudwatch.Region)
 		v.Set("logging.cloudwatch.secret", clowderConfig.Logging.Cloudwatch.SecretAccessKey)
+		// TODO Delete the block below when the below PR is merged
+		// See: https://github.com/RedHatInsights/clowder/pull/627
+		if clowderConfig.Logging.Type == "" {
```
cryptomilk, line 295: Maybe also check if `len(clowderConfig.Logging.Cloudwatch.SecretAccessKey) > 0`?

tiran: How about we only check for secret access keys? That way the code will still work when https://github.com/RedHatInsights/clowder/pull/627 has been merged.

avisiedo: IMHO I don't know if another logging providers will coexists at some moment in the future; the guard `Type == ""` avoid to overwrite if some value exists at that moment, the check `len(clowderConfig.Logging.Cloudwatch.SecretAccessKey) > 0` will additionally set the value when we have some information for cloudwatch.

The current change accomplish:
- avoid overwrite value if some content exists.
- avoid to set the value if maybe cloudwatch object exists but the contents are empty.
- this let to set `"cloudwatch"` value meanwhile the PR is merged and will let the code works after merging it.

cryptomilk: I think the workaround is good as it is right now. In future you might have cloudwatch secrets but logging.type configured to something else and want to react on this.

cryptomilk (approved):
LGTM

internal/config/config.go
```diff
@@ -290,6 +290,11 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 		v.Set("logging.cloudwatch.group", clowderConfig.Logging.Cloudwatch.LogGroup)
 		v.Set("logging.cloudwatch.region", clowderConfig.Logging.Cloudwatch.Region)
 		v.Set("logging.cloudwatch.secret", clowderConfig.Logging.Cloudwatch.SecretAccessKey)
+		// TODO Delete the block below when the below PR is merged
+		// See: https://github.com/RedHatInsights/clowder/pull/627
+		if clowderConfig.Logging.Type == "" && len(clowderConfig.Logging.Cloudwatch.SecretAccessKey) > 0 {
+			v.Set("logging.type", "cloudwatch")
+		}
```
tiran, line 297: Let's future proof the check (untested code)

```suggestion
		// TODO Remove check for empty string after clowder has been fixed.
		// See: https://github.com/RedHatInsights/clowder/pull/627
		if (clowderConfig.Logging.Type == "" || clowderConfig.Logging.Type == "cloudwatch") && len(clowderConfig.Logging.Cloudwatch.SecretAccessKey) > 0 {
			v.Set("logging.type", "cloudwatch")
		}
```

cryptomilk: Why set the logging type to cloudwatch if it already set? You might want to check the code with more context.

cryptomilk:
I've talked to the author of the clowder fix (psav). They discovered two years ago, that they always set logging.type to "" instead of "cloudwatch". However a lot of projects already check for "" as the cloudwatch case so they didn't commit to fix it. He said all projects would have to change.

However it is a bug and should be fixed, but in the meantime a lot more projects use cloudwatch and the number of projects implementing a workaround increases. So the question is if them will ever fix it or not. If they don't fix it, they should check all the documentations and remove logging.type = "cloudwatch".

I think that the current implementation is the best workaround whatever they decide :-)

tiran (commented):
LGTM, please rebase and squash all commits into a single commit.

avisiedo:
My apologize, I realized about squash the commits when already pressed the merge button.

PR #164, opened 2024-03-21
fix(HMS-3805): add mixed enforcement

avisiedo:
This change allow to define routes for mixed enforcement for the routes that could be reached out by users or systems.

### DIFF
--- a/internal/infrastructure/middleware/enforce_identity.go
+++ b/internal/infrastructure/middleware/enforce_identity.go
@@ -57,7 +57,7 @@ func EnforceUserPredicate(data *identity.XRHID) error {
 		return fmt.Errorf("'data' cannot be nil")
 	}
 	if data.Identity.Type != "User" {
-		return fmt.Errorf("'Identity.Type' is not 'User'")
+		return fmt.Errorf("'Identity.Type=%s' is not 'User'", data.Identity.Type)
 	}
 	if !data.Identity.User.Active {
 		return fmt.Errorf("'Identity.User.Active' is not true")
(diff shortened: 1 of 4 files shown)


tiran (changes requested):
I don't particularly like the current approach. Before we had two lookup tables and check functions. Now we have three. It's code duplication and three places to check if one wants to understand access permissions

I would rather do something like this:

```golang
type enforceRoute struct {
	Method string
	Path   string
	System bool
    User bool
}

and the define permissions like this:
```golang
{"GET", "/api/idmsvc/v1/domains/:uuid", true, true},
```

or maybe use a bit map:

```golang
USER_ACCESS = 1
SYSTEM_ACCESS = 2
ALL_ACCESS = USER_ACCESS | SYSTEM_ACCESS
```

```golang
type enforceRoute struct {
	Method string
	Path   string
	Access uint
}

{"GET", "/api/idmsvc/v1/domains/:uuid", ALL_ACCESS}

if route.Access & USER_ACCESS == USER_ACCESS { ... }
```

avisiedo:
I agree some refactor is necessary, but it was not my intention to make that refactor on this PR. Thanks for the observation.

frasertweedale:
Yeah I agree we can refactor later.  In fact, we already have a ticket that proposes a nicer way: https://issues.redhat.com/browse/HMS-2459 .

frasertweedale (approved):

tiran:
I also noticed that the change gives a domain too much access. A domain should only be able to view its own data and not be able to access any other domain.

frasertweedale:
> I also noticed that the change gives a domain too much access. A domain should only be able to view its own data and not be able to access any other domain.

The lookup filters by org ID, and therefore 404s if the requested domain exists in a different org.

For the middleware to enforce it, we would have to do the lookup in the middleware.  In any case, I don't think the current behaviour is wrong.  But if I am missing something please put me straight!

frasertweedale:
There were no further comment or objections, so merging this now.

PR #175, opened 2024-04-02
test: add stage scripts

avisiedo:
Add a set of scripts to make life easier when
dealing with stage environment.

### DIFF
--- a/test/scripts/stage-domains-delete.sh
+++ b/test/scripts/stage-domains-delete.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -eo pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/stage.inc"
+
+UUID="$1"
+[ "${UUID}" != "" ] || error "UUID is empty"
+
+unset X_RH_IDENTITY
+unset X_RH_FAKE_IDENTITY
+unset X_RH_IDM_REGISTRATION_TOKEN
+X_RH_IDM_VERSION="$IDM_VERSION"
+export X_RH_IDM_VERSION
+
+exec "${REPOBASEDIR}/scripts/curl.sh" -i -X DELETE "${BASE_URL}/domains/${UUID}"
--- a/test/scripts/stage-domains-register.sh
+++ b/test/scripts/stage-domains-register.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -eo pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/stage.inc"
+
+TOKEN="$1"
+[ "${TOKEN}" != "" ] || error "TOKEN is empty"
+
+unset X_RH_IDENTITY
+unset X_RH_FAKE_IDENTITY
+export X_RH_IDM_REGISTRATION_TOKEN="${TOKEN}"
+X_RH_IDM_VERSION="$IDM_VERSION"
+export X_RH_IDM_VERSION
+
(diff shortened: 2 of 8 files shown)


test/scripts/stage.inc
```diff
@@ -0,0 +1,44 @@
...
+#
+# See: ./scripts/curl.sh
+#
+source "$(dirname "${BASH_SOURCE[0]}")/common.inc"
+
+NAMESPACE="idmsvc-stage"
+export NAMESPACE
+
+if [ "${https_proxy}" == "" ] && [ "${http_proxy}" == "" ]; then
+    error "https_proxy and http_proxy are empty: check your proxy configuration"
```
pvoborni, line 16: Using a proxy might not be always needed - it depends on where the script is run. E.g. it is needed when using vpn.

Could the script advise what to set when it will fail because of missing proxy info?

test/scripts/stage-domains-register.sh
```diff
@@ -0,0 +1,15 @@
...
+TOKEN="$1"
+[ "${TOKEN}" != "" ] || error "TOKEN is empty"
+
+unset X_RH_IDENTITY
+unset X_RH_FAKE_IDENTITY
+export X_RH_IDM_REGISTRATION_TOKEN="${TOKEN}"
+X_RH_IDM_VERSION="$IDM_VERSION"
+export X_RH_IDM_VERSION
+
+exec "${REPOBASEDIR}/scripts/curl.sh" -i -X POST -d @<(sed -e 's/{{subscription_manager_id}}/6f324116-b3d2-11ed-8a37-482ae3863d30/g' < "${REPOBASEDIR}/test/data/http/register-rhel-idm-domain.json") "${BASE_URL}/domains"
```
pvoborni, line 15: Should it work?  Resp. are we able to make it work with this "fake script".   The `stage.inc` does auth for `user` or for  `service account` but the POST method expects a `system`.   So when I tried with default settings I got 401.

pvoborni (commented):
I didn't test `delete` and `patch` due to the issues with register and I did not want to remove the current domain I have in stage.

test/scripts/stage-domains-register.sh
avisiedo: thanks! good catch! no! this should not work! I have to remove this one!

test/scripts/stage.inc
```diff
@@ -0,0 +1,44 @@
+#
+# Include for common parts for ephemeral environment shared between all the scripts
```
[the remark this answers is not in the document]
avisiedo: thanks! yes it is! sorry!

test/scripts/stage-domains-delete.sh
```diff
@@ -0,0 +1,15 @@
...
+set -eo pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/stage.inc"
+
+UUID="$1"
+[ "${UUID}" != "" ] || error "UUID is empty"
+
+unset X_RH_IDENTITY
+unset X_RH_FAKE_IDENTITY
+export X_RH_IDM_REGISTRATION_TOKEN="${TOKEN}"
```
pvoborni, line 11: The token part is not needed in delete.

pvoborni (approved):
I'll approve as the `list`, `token`, `patch`, `delete` scripts work and thus it creates value.

The `register` script doesn't work as it doesn't use system identity, but this might be better to fix in smaller separate PR, e.g. with using real certs from some stage systems on each devel's machine.

PR #163, opened 2024-03-21
test(HMS-3527): provide rbac mock support

avisiedo:
WIP

The change provide basic mocking for rbac service related with what we need for our smoke tests, and integrate the admin profile for the current smoke tests.

TODO:
- [x] Create initial infrastructure to run rbac mock service, and integrate changes.
- [x] Add interface to communicate between the smoke test and the rbac mock.
- [ ] (80%) Add more unit tests.
- [ ] (80%) Integrate with current smoke tests by using the admin profile.
- [ ] Stabilize the changes and tests.
- [ ] reorg commit

PR #178, opened 2024-04-03
refactor(HMS-3527): mock rbac standalone

avisiedo:
Refactor the rbac mock to get a standalone service and
update local infrastructure to run it as a container.

The container is available at: [quay.io/podengo/mock-rbac:1.0.0](https://quay.io/repository/podengo/mock-rbac?tab=tags).

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/163


.vscode/launch.json
```diff
@@ -14,6 +14,14 @@
             "args": [],
             "env": {
                 "CONFIG_PATH": "${workspaceFolder}/configs",
+                "ENV_NAME": "local",
```
tiran, line 17: The variable name `ENV_NAME` is rather generic. What are other possible values? Is it just for RBAC testing or is it used for other settings, too?

build/mock-rbac/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+##
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
+LABEL idmsvc-backend=builder
+# https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant
+ENV OPENSSL_FORCE_FIPS_MODE=1
+WORKDIR /go/src/app
+COPY . .
+USER 0
+RUN git log -1
+RUN make get-deps bin/mock-rbac
```
tiran, line 12: Do we need `get-deps`? Does the container build work without it?
```suggestion
RUN make bin/mock-rbac
```

build/mock-rbac/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3-1552
+LABEL idmsvc-backend=backend
+# https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant
+ENV OPENSSL_FORCE_FIPS_MODE=1
+RUN mkdir -p /opt/bin /opt/bin/scripts/db /opt/bin/configs
+WORKDIR /opt/bin
+COPY --from=builder /go/src/app/bin/mock-rbac ./
+COPY scripts/db/migrations /opt/bin/scripts/db/migrations
```
tiran, line 22: Why do we need DB migrations in the mock container?

internal/config/config.go
```diff
@@ -328,6 +328,17 @@ func Load(cfg *Config) *viper.Viper {
...
+		reportError(err)
+		panic("Invalid configuration")
+	}
+
+	sec, err := secrets.NewAppSecrets(cfg.Application.MainSecret)
+	if err != nil {
+		panic(err)
+	}
+	cfg.Secrets = *sec
+
```
tiran, line 341: As discussed in the other PR, this section should not be moved.

tiran (changes requested):

build/mock-rbac/Dockerfile
avisiedo: thanks! we don't need it! "copy&paste"

.vscode/launch.json
avisiedo: I was wrong! I have updated the PR with this change! thanks!

internal/config/config.go
avisiedo: thanks! Restoring!

frasertweedale (approved):
Tested locally, tests are passing after the config change (which is mentioned in comments and in the commit message - thanks!).  So, let's merge it.

PR #183, opened 2024-04-05
fix(HMS-3914): hardener container image build

avisiedo:
Use a "ignore by default" at .dockerignore and
explicitly add exceptions for the content to copy
to the container.


avisiedo:
closing; duplicated and added better solution at https://github.com/podengo-project/idmsvc-backend/pull/187

PR #187, opened 2024-04-10
fix(HMS-3914): harden dockerfile

avisiedo:
Add a policy to ignore by default from .dockerignore file. As .git is not copied, the printing out the last commit is made from the container-build rule. For internal CI/CD we can track the repository related state by the image tag which is the short hash of the repository.

Credits on: insights-rbac contributors

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,7 @@
+.podman
+.docker
+.kube
+
 .DS_Store
 .cicd_bootstrap.sh
 secrets


tiran:
I fail to understand what problem this is solving. How can we ever leak credentials when the final container image only contains the binaries, migration scripts, and example config? AFAIK the intermediate build container will never ever leave the build machine.

Also we require the `.git` directory at build time to include VCS information in the binary. The `RUN git log` line is there to verify that the `.git` directory is available and accessible.

PR #195, opened 2024-04-23
fix(HMS-4004): tools/go.mod dependency on go.tmz.dev/musttag

avisiedo:
This change add a replace directive to fix the dependency
on `go.tmz.dev/musttag`.

### DIFF
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -243,3 +243,5 @@ require (
 )
 
 exclude github.com/mitchellh/osext v0.0.0-20151018003038-5e2d6d41470f
+
+replace go.tmz.dev/musttag v0.7.2 => go-simpler.org/musttag v0.7.2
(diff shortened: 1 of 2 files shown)


pvoborni (approved):
LGTM.  

This is fixing the issue I encountered when building the tools (`make install-tools`) without a need to bump `golangci-lint` that would require a bump of used golang to `1.21`.

avisiedo:
I think I pushed after the approved, and it is required again. I thought the `Pair-Programming-With` was recognized by github, but it only recognizes `Co-authored-by: ...` on the commit tail.

pvoborni (approved):
Re-approving, but I think I'm not on approval list to be able to fully "approve" the PR.  From my PoV, this is not a big issue as I'm new to the project.  But it needs e.g. @frasertweedale

PR #198, opened 2024-05-03
fix(HMS-4061): fix auto enrollment

avisiedo:
When the register operation happens, it is set
auto-enrollment to false by default. This is for
the `POST /domains` operation.

Later the user decide if it is enable or disabled
with a `PATCH /domains/:domain_id` operation, which
match with clicking on the **Finish** button on the
registration domain wizard.

It can be checked by:

```shell
# Start the backend
make compose-clean clean build commpose-up run
# From another terminal run the below
# Generate a token
./test/scripts/local-domain-token.sh
TOKEN="..."  # the value returned for it
# Register the example domain
./test/scripts/local-domain-register "$TOKEN"
# Check the database
make db-cli
select * from domains;
# You will see the field for auto_enrollment_enabled is f
# which means it is disabled
```

### DIFF
--- a/internal/usecase/interactor/domain_interactor.go
+++ b/internal/usecase/interactor/domain_interactor.go
@@ -178,8 +178,8 @@ func (i domainInteractor) Register(domainRegKey []byte, xrhid *identity.XRHID, p
 	// default title is domain name
 	domain.Title = pointy.String(body.DomainName)
 	domain.Description = pointy.String("")
-	// new domains are enabled by default
-	domain.AutoEnrollmentEnabled = pointy.Bool(true)
+	// new domains are disabled by default
+	domain.AutoEnrollmentEnabled = pointy.Bool(false)
 
 	return orgID, clientVersion, domain, nil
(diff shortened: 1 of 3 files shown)


avisiedo:
I need to double check the UI before remove the draft, and be sure the ipa-hcc pipeline don't get broken by this change.

internal/usecase/interactor/domain_interactor.go
```diff
@@ -179,7 +179,7 @@ func (i domainInteractor) Register(domainRegKey []byte, xrhid *identity.XRHID, p
 	domain.Title = pointy.String(body.DomainName)
 	domain.Description = pointy.String("")
 	// new domains are enabled by default
```
frasertweedale, line 181: please update the comment :)

avisiedo: :+1: thanks! updated!

frasertweedale (approved):
Tested, works as advertised.

PR #220, opened 2024-06-05
feat(HMS-3528): enable rbac by default

avisiedo:
This change enable RBAC by default, this means the below:

- rbac is enabled by default when we run locally the service by `make run`. See: secrets/private.mk to avoid it is override the value.
- rbac is enabled by default when we deploy in dev cluster by `make ephemeral-deploy` or `make ephemeral-build-deploy`.
- rbac is enabled by default when we deploy in stage, unless the parameter APP_ENABLE_RBAC is override on deployment time.

### DIFF
--- a/configs/bonfire.example.yaml
+++ b/configs/bonfire.example.yaml
@@ -33,7 +33,7 @@ apps:
           APP_ACCEPT_X_RH_FAKE_IDENTITY: "true"
           APP_VALIDATE_API: "true"
           APP_TOKEN_EXPIRATION_SECONDS: "7200"
-          APP_ENABLE_RBAC: "true"
+          # APP_ENABLE_RBAC: "true"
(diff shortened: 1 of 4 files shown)


avisiedo:
point (1) currently require the developer set the value for `enable_rbac: true` at `configs/config.yaml` file, or APP_ENABLE_RBAC := true` at secrets/private.mk; with this change the user could remove that (but still could disable it when needed).

point (2) allow that the default value will be True, so make ephemeral-deploy will require less burden at `./scripts/mk/ephemeral.mk`; and it removes additional burden on configuration (configs/bonfire.yaml).

point (3) allows that the default value will be Tue on deployment time; it does not remove the feature flag (deployment parameter), but if it is not specified, the default value will be true. Related in part to point (2); this will allow to just run `bunfore deploy idmsvc` and rbac will be enabled.

pvoborni (approved):
LGTM

PR #221, opened 2024-06-07
fix(HMS-4274): rbac calls for system endpoints

avisiedo:
System calls are failing; this change fix the situation calling to rbac endpoints when it is required.

The tests for the System endpoints has been extracted in its own test suite and additional endpoints for the smoke tests has been added there. A new builder is added for HostConfJwk instances, it prints more logs for the HostConf endpoint handler when an error happens.

Additionally
- a confusing error message when rbac mock close connection is removed.
- it fix rbac mapping created for the unit tests of the rbac middleware.
- add missed `/api/idmsvc/v1/signing_keys` at the system enforcement routes.
- include suggestions from @pvoborni .
- fix a situation to delete keys only when the slice has items.

https://issues.redhat.com/browse/HMS-4274
https://issues.redhat.com/browse/HMS-3525

### DIFF
--- a/internal/infrastructure/middleware/rbac_test.go
+++ b/internal/infrastructure/middleware/rbac_test.go
@@ -31,17 +31,13 @@ func helperRbacCreateMapping() rbac_data.RBACMap {
 	service := rbac_data.RBACService("idmsvc")
 	resourceToken := rbac_data.RBACResource("token")
 	resourceDomains := rbac_data.RBACResource("domains")
-	resourceSigningKeys := rbac_data.RBACResource("signing_keys")
-	resourceHostConf := rbac_data.RBACResource("host_conf")
+	// This map fit the one at internal/infrastructure/router/rbac.yaml
 	data := rbac_data.NewRBACMapBuilder().
 		Add("/domains/token", http.MethodPost, rbac_data.NewRbacPermission(service, resourceToken, rbac_data.RbacVerbCreate)).
 		Add("/domains", http.MethodGet, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbRead)).
 		Add("/domains/:uuid", http.MethodGet, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbRead)).
-		Add("/domains/:uuid", http.MethodPost, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbCreate)).
 		Add("/domains/:uuid", http.MethodPatch, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbUpdate)).
 		Add("/domains/:uuid", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbDelete)).
-		Add("/signing_keys", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceSigningKeys, rbac_data.RbacVerbExecute)).
-		Add("/host-conf/:inventory_id/:fqdn", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceHostConf, rbac_data.RbacVerbExecute)).
 		Build()
 	return data
 }
--- a/internal/infrastructure/router/public.go
+++ b/internal/infrastructure/router/public.go
@@ -75,11 +75,12 @@ func newRbacSkipper(service string) echo_middleware.Skipper {
 			cc middleware.DomainContextInterface
 			ok bool
 		)
+		openapiPath := "/api/" + service + "/v1/openapi.json"
 		ctx := c.Request().Context()
 		routePath := c.Path()
 		// The access to the openapi specification is public
-		if routePath == "/api/"+service+"/v1/openapi.json" {
-			slog.DebugContext(ctx, "Skipping "+routePath)
+		if routePath == openapiPath {
+			slog.DebugContext(ctx, "route is '"+openapiPath+"'")
 			return true
 		}
 		if cc, ok = c.(middleware.DomainContextInterface); !ok {
@@ -87,12 +88,7 @@ func newRbacSkipper(service string) echo_middleware.Skipper {
 			return false
 		}
 		if cc.XRHID().Identity.Type == "System" {
-			for i := range systemEnforceRoutes {
-				if routePath == systemEnforceRoutes[i].Path {
-					slog.DebugContext(ctx, "Skipping "+routePath+" for the System with CN="+cc.XRHID().Identity.System.CommonName)
-					return true
-				}
-			}
+			return true
 		}
 		return false
 	}
--- a/internal/infrastructure/router/public.go
+++ b/internal/infrastructure/router/public.go
@@ -3,6 +3,7 @@ package router
 import (
 	_ "embed"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/labstack/echo/v4"
@@ -37,6 +38,7 @@ var systemEnforceRoutes = []enforceRoute{
 	{"POST", "/api/idmsvc/v1/domains"},
 	{"PUT", "/api/idmsvc/v1/domains/:uuid"},
 	{"POST", "/api/idmsvc/v1/host-conf/:inventory_id/:fqdn"},
+	{"GET", "/api/idmsvc/v1/signing_keys"},
 }
 
 var mixedEnforceRoutes = []enforceRoute{
@@ -69,7 +71,30 @@ func newRbacSkipper(service string) echo_middleware.Skipper {
 		panic("service is an empty string")
 	}
 	return func(c echo.Context) bool {
-		return c.Path() == "/api/"+service+"/v1/openapi.json"
+		var (
+			cc middleware.DomainContextInterface
+			ok bool
+		)
+		ctx := c.Request().Context()
+		routePath := c.Path()
+		// The access to the openapi specification is public
+		if routePath == "/api/"+service+"/v1/openapi.json" {
+			slog.DebugContext(ctx, "Skipping "+routePath)
+			return true
+		}
+		if cc, ok = c.(middleware.DomainContextInterface); !ok {
+			slog.WarnContext(ctx, "'c' is not a DomainContextInterface")
+			return false
+		}
+		if cc.XRHID().Identity.Type == "System" {
+			for i := range systemEnforceRoutes {
+				if routePath == systemEnforceRoutes[i].Path {
+					slog.DebugContext(ctx, "Skipping "+routePath+" for the System with CN="+cc.XRHID().Identity.System.CommonName)
+					return true
+				}
+			}
+		}
+		return false
 	}
 }
 
--- a/internal/test/smoke/rbac_permission_test.go
+++ b/internal/test/smoke/rbac_permission_test.go
@@ -218,31 +218,6 @@ func (s *SuiteRbacPermission) helperCommonAdmin() []TestCasePermission {
 			Then:     s.doTestDomainList,
 			Expected: http.StatusOK,
 		},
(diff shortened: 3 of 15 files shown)


internal/infrastructure/router/public.go
```diff
@@ -69,7 +70,27 @@ func newRbacSkipper(service string) echo_middleware.Skipper {
...
+		if c.Path() == "/api/"+service+"/v1/openapi.json" {
+			slog.DebugContext(ctx, "Skipping "+c.Path())
+			return true
+		}
+		cc, ok := c.(middleware.DomainContextInterface)
+		if !ok {
+			slog.WarnContext(ctx, "'c' is not a DomainContextInterface")
+			return false
+		}
+		if cc.XRHID().Identity.Type == "System" {
```
pvoborni, line 84: Is the routes check needed? Given that `enforce identity` middlewares check if the right identity type is present for various routes, we could use something like this:
```go
		cc, ok := c.(middleware.DomainContextInterface)
		if !ok {
			slog.DebugContext(ctx, "RBAC: not a DomainContextInterface")
			return true
		}

		// RBAC is applicable only for path that have RH ID of type User
		identity := cc.XRHID()
		if identity == nil {
			slog.DebugContext(ctx, "RBAC: XRHID is not set")
			return true
		}
		if identity.Identity.Type != "User" {
			slog.DebugContext(ctx, "RBAC: XRHID is not a User")
			return true
		}
```

Btw, I' m not logging with word "skipping" and the "path" as it is logged in the `rbac.go`. 

Side-note, I wonder what `type` has service account. Looking into https://github.com/RedHatInsights/platform-go-middlewares/blob/master/identity/identity.go#L55 , there seems to be 5 structs that look as identities. 

If service account has a different type, then `userIdentityMiddleware` and `mixedIdentityMiddleware` would need to be enhanced with a new predicate and by code here with an additional type. Or it can be

```
		if identity.Identity.Type == "System" {
			slog.DebugContext(ctx, "RBAC: XRHID has type 'System'")
			return true
		}
```

internal/test/smoke/rbac_permission_test.go
```diff
@@ -239,13 +239,14 @@ func (s *SuiteRbacPermission) TestReadPermission() {
 			Name:     "Test idmsvc:domain:create",
 			Given:    s.prepareDomainIpaCreate,
 			Then:     s.doTestDomainIpaCreate,
-			Expected: http.StatusUnauthorized,
+			Expected: http.StatusCreated,
```
pvoborni, line 242: Is it correct? When user have only read permissions that doing API request to create domain should IMO fail.  It can be different for IPA server doing the request though - a system account.

internal/test/smoke/rbac_permission_test.go
```diff
@@ -239,13 +239,14 @@ func (s *SuiteRbacPermission) TestReadPermission() {
...
 		{
-			Name:     "Test Update Agent idmsvc:domain:update",
-			Given:    s.prepareDomainIpa,
-			Then:     s.doTestDomainIpaUpdate,
-			Expected: http.StatusUnauthorized,
+			Name:  "Test Update Agent idmsvc:domain:update",
+			Given: s.prepareDomainIpa,
+			Then:  s.doTestDomainIpaUpdate,
+			// System requests are identified by its certificate
+			Expected: http.StatusOK,
```
pvoborni, line 249: Same here. Also I don't understand much the comment `// System requests are identified by its certificate` - how it is related to this test.

pvoborni (commented):
Added some questions.

internal/infrastructure/router/public.go
avisiedo: Thanks for the suggestion.

Yes, enforcement should warranty that `XRHID` is valid and store the unserialized value in the context. I would add the change :+1:

About the route check, I am afraid it is necessary.

avisiedo: I step back; you are right, the route check is not necessary at the rbac skipper. I will add it :+1:  thanks

internal/test/smoke/rbac_permission_test.go
avisiedo: Thanks! It is wrong how I structured the table; but the expected is right; that endpoint would not be reached out by a user; it is enforced the type of XRHID for this endpoint to be `System`. :+1:

internal/infrastructure/router/public.go
```diff
@@ -69,7 +70,27 @@ func newRbacSkipper(service string) echo_middleware.Skipper {
...
+		ctx := c.Request().Context()
+		// The access to the openapi specification is public
+		if c.Path() == "/api/"+service+"/v1/openapi.json" {
+			slog.DebugContext(ctx, "Skipping "+c.Path())
+			return true
+		}
+		cc, ok := c.(middleware.DomainContextInterface)
+		if !ok {
+			slog.WarnContext(ctx, "'c' is not a DomainContextInterface")
+			return false
```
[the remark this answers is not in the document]
avisiedo: thanks for the suggestion! it can happen in a bug! indeed this PR fix a situation detected when adding the smoke test for the System endpoints. I suggest to keep the return as `false`, as a protection when in the future the API changes or extends.

internal/test/smoke/rbac_permission_test.go
avisiedo: thanks! I have restructured the table, but that set of tests make sense in a different suite dedicated to the System endpoints. I will refactor! :+1:

internal/usecase/repository/host_repository.go
```diff
@@ -43,7 +43,8 @@ func (r *hostRepository) MatchDomain(db *gorm.DB, options *interactor.HostConfOp
 		Joins("left join ipas on domains.id = ipas.id").
 		Where("domains.org_id = ?", options.OrgId)
 	if options.DomainId != nil {
-		tx = tx.Where("domains.domain_uuid = ?", options.DomainId.String())
+		domainUUID := options.DomainId.String()
+		tx = tx.Where("domains.domain_uuid = ?", domainUUID)
```
avisiedo, line 47: I have divided this for better debugging, so I can see at the domainUUID the string representation.

internal/usecase/repository/hostconf_jwk_repository.go
```diff
@@ -105,12 +105,14 @@ func (r *hostconfJwkRepository) PurgeExpiredJWKs(db *gorm.DB) (hcjwks []model.Ho
...
-		slog.Error(err.Error())
-		return nil, err
+	if len(hcjwks) > 0 {
+		if err = db.
+			Unscoped(). // do not use GORM's soft delete for purging
+			Delete(&hcjwks).
+			Error; err != nil {
+			slog.Error(err.Error())
+			return nil, err
+		}
```
avisiedo, line 115: Meanwhile debugging I got some situations when no keys exists and executing the `Delete` operation with an empty slice, so only execute `Delete` when there is something to delete.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -0,0 +1,181 @@
...
+	domainID := s.domain.DomainId.String()
+	res, err := s.UpdateDomainWithResponse(
+		domainID,
+		builder_api.NewUpdateDomainAgent("test.example").
+			WithHCCUpdate(true).
+			WithDomainRhelIdm(*builder_api.NewRhelIdmDomain("test.example").
+				WithServers([]public.DomainIpaServer{}).
+				AddServer(
+					builder_api.NewDomainIpaServer("1.test.example").
+						WithHccUpdateServer(true).
```
avisiedo, line 140: This line fixed the last flaky test.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -0,0 +1,181 @@
...
+	s.RbacMock.SetPermissions(mock_rbac.Profiles[mock_rbac.ProfileDomainAdmin])
+	// Create a token to register a domain
+	s.token, err = s.CreateToken()
+	require.NoError(t, err)
+	require.NotNil(t, s.token)
+	require.NotEqual(t, "", s.token.DomainToken)
+
+	// This operation set AutoEnrollmentEnabled = False whatever
+	// is the value we indicate here; we have to PATCH in a second
+	// operation
```
pvoborni, line 165: Is this comment needed here as there is not second operation done here?

internal/infrastructure/middleware/rbac_test.go
```diff
@@ -40,8 +40,8 @@ func helperRbacCreateMapping() rbac_data.RBACMap {
 		Add("/domains/:uuid", http.MethodPost, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbCreate)).
 		Add("/domains/:uuid", http.MethodPatch, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbUpdate)).
 		Add("/domains/:uuid", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceDomains, rbac_data.RbacVerbDelete)).
-		Add("/signing_keys", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceSigningKeys, rbac_data.RbacVerbExecute)).
-		Add("/host-conf/:inventory_id/:fqdn", http.MethodDelete, rbac_data.NewRbacPermission(service, resourceHostConf, rbac_data.RbacVerbExecute)).
+		Add("/signing_keys", http.MethodPost, rbac_data.NewRbacPermission(service, resourceSigningKeys, rbac_data.RbacVerbExecute)).
```
pvoborni, line 43: I don't know what these 2 lines mean / what is their impact.  When commented out, the tests still pass.

pvoborni (commented):
Hi, mostly looks good to me. I've added some talking points, most of them is not something to address here.

internal/infrastructure/middleware/rbac_test.go
pvoborni: Discussion: these 2 lines should be removed + POST of `domains/:uuid`

pvoborni (approved):
LGTM

internal/test/smoke/suite_base_test.go
```diff
@@ -239,11 +239,14 @@ func (s *SuiteBase) RegisterIpaDomain(token string, domain *public.Domain) (*pub
...
-func (s *SuiteBase) ReadDomainWithResponse(domainID uuid.UUID) (*http.Response, error) {
+func (s *SuiteBase) readDomainWithResponse(xrhid *identity.XRHID, domainID uuid.UUID) (*http.Response, error) {
+	if xrhid == nil {
+		panic("'xrhid' is nil")
+	}
 	hdr := http.Header{}
 	url := s.DefaultPublicBaseURL() + "/domains/" + domainID.String()
 	method := http.MethodGet
-	s.addXRHIDHeader(&hdr, &s.UserXRHID)
+	s.addXRHIDHeader(&hdr, xrhid)
```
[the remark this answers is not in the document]
avisiedo: Follow up on ticket: https://issues.redhat.com/browse/HMS-4280

internal/test/smoke/suite_base_test.go
```diff
@@ -445,6 +485,77 @@ func (s *SuiteBase) PatchDomain(domainID string, domain *public.UpdateDomainUser
...
+		body,
+	)
+	return resp, err
+}
+
+func (s *SuiteBase) SystemHostConf(inventoryID string, fqdn string, hostconf *public.HostConf) (*public.HostConfResponse, error) {
+	url := s.DefaultPublicBaseURL() + "/host-conf/" + inventoryID + "/" + fqdn
+	resp, err := s.SystemHostConfWithResponse(inventoryID, fqdn, hostconf)
+
+	if resp.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
avisiedo: Follow up on ticket: https://issues.redhat.com/browse/HMS-4280

PR #232, opened 2024-06-15
refactor(HMS-3968): add `As` semantics

avisiedo:
This PR add `As` semantics so it can be used for
set the RBACProfile or XRHIDProfile; we can provide
several in the list of arguments, and they are applied
from left to right; the last value is the one that take
effect. More details below:

About XRHID:

- As(XRHIDServiceAccount): it indicates that following
  API operations will use a service account XRHID.
- As(XRHIDUser): it indicates that following API
  operations will use a user XRHID.
- As(XRHIDSystem): it indicates that following API
  operations will use a System XRHID.
- As(XRHIDNone): it indicates that following API operations
  will not add any XRHID header. It forces to delete the
  header when the request is made.
- As(XRHIDNothing): it indicates that following API
  operations will not add / delete any XRHID header.

About RBAC:

- As(RBACSuperAdmin): tell to the rbac mock to return super admin
  permissions for the following requests.
- As(RBACAdmin): tell to the rbac mock to return admin permissions
  for the following requests.
- As(RBACReadOnly): tell to the rbac mock to return read only
  permissions for the following requests.
- As(RBACNoPerms): tell to the rbac mock to return empty list
  of permissions for the following requests.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/235

### DIFF
--- a/internal/test/builder/api/builder_service_account_xrhid.go
+++ b/internal/test/builder/api/builder_service_account_xrhid.go
@@ -1,0 +1,94 @@
+package api
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/podengo-project/idmsvc-backend/internal/test/builder/helper"
+	builder_helper "github.com/podengo-project/idmsvc-backend/internal/test/builder/helper"
+	identity "github.com/redhatinsights/platform-go-middlewares/v2/identity"
+)
+
+type ServiceAccountXRHID interface {
+	Build() identity.XRHID
+	WithAccountNumber(value string) ServiceAccountXRHID
+	WithEmployeeAccountNumber(value string) ServiceAccountXRHID
+	WithOrgID(value string) ServiceAccountXRHID
+	WithAuthType(value string) ServiceAccountXRHID
+
+	WithClientID(value string) ServiceAccountXRHID
+	WithUsername(value string) ServiceAccountXRHID
+}
+
+type serviceAccountXRHID identity.XRHID
+
+// NewServiceAccountXRHID create a new builder to get a XRHID for a user
+func NewServiceAccountXRHID() ServiceAccountXRHID {
+	var orgID = strconv.Itoa(int(builder_helper.GenRandNum(1, 100000)))
+	// See: https://github.com/RedHatInsights/identity-schemas/blob/main/3scale/identities/service-account.json
+	return &serviceAccountXRHID{
+		Identity: identity.Identity{
+			AccountNumber:         strconv.Itoa(int(builder_helper.GenRandNum(1, 100000))),
+			EmployeeAccountNumber: strconv.Itoa(int(builder_helper.GenRandNum(1, 100000))),
+			OrgID:                 orgID,
+			Type:                  "ServiceAccount",
+			AuthType:              "jwt-auth",
+			Internal: identity.Internal{
+				OrgID:    orgID,
+				AuthTime: float32(time.Now().Sub(time.Time{})),
+			},
+			ServiceAccount: &identity.ServiceAccount{
+				ClientId: uuid.NewString(),
+				Username: helper.GenRandUsername(),
+			},
+		},
+	}
+}
+
+func (b *serviceAccountXRHID) Build() identity.XRHID {
+	return identity.XRHID(*b)
+}
+
+func (b *serviceAccountXRHID) WithOrgID(value string) ServiceAccountXRHID {
+	b.Identity.OrgID = value
+	b.Identity.Internal.OrgID = value
+	return b
+}
+
+func (b *serviceAccountXRHID) WithAuthType(value string) ServiceAccountXRHID {
+	switch value {
+	case authType[0]:
+		b.Identity.AuthType = value
+		return b
+	case authType[1]:
+		b.Identity.AuthType = value
+		return b
+	default:
+		panic(fmt.Sprintf("value='%s' is not valid", value))
+	}
+}
+
+func (b *serviceAccountXRHID) WithAccountNumber(value string) ServiceAccountXRHID {
+	b.Identity.AccountNumber = value
+	return b
+}
+
+func (b *serviceAccountXRHID) WithEmployeeAccountNumber(value string) ServiceAccountXRHID {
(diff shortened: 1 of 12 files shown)


internal/test/builder/api/builder_service_account_xrhid.go
```diff
@@ -0,0 +1,94 @@
...
+	WithClientID(value string) ServiceAcccountXRHID
+	WithUsername(value string) ServiceAcccountXRHID
+}
+
+type serviceAcccountXRHID identity.XRHID
+
+// NewServiceAccountXRHID create a new builder to get a XRHID for a user
+func NewServiceAccountXRHID() ServiceAcccountXRHID {
+	var orgID = strconv.Itoa(int(builder_helper.GenRandNum(1, 100000)))
+	// See: https://github.com/coderbydesign/identity-schemas/blob/add-validator/3scale/schema.json
```
pvoborni, line 30: The link doesn't work. Should it be https://github.com/RedHatInsights/identity-schemas/blob/main/3scale/identities/service-account.json as in the previous PR?

avisiedo: Thanks! you are right! updated!

internal/test/smoke/suite_base_test.go
```diff
@@ -592,6 +633,7 @@ func (s *SuiteBase) RunTestCase(testCase *TestCase) {
 	}
 
 	// WHEN
+	s.As(XRHIDProfile(testCase.Given.XRHIDProfile))
```
pvoborni, line 636: Is there is a reason to use the type conversion here? It looks like that `mate test` works also with `s.As(testCase.Given.XRHIDProfile)`.

internal/test/smoke/rbac_permission_test.go
```diff
@@ -109,20 +114,25 @@ func (s *SuiteRbacPermission) doTestDomainList(t *testing.T) int {
 	return res.StatusCode
 }
 
-func (s *SuiteRbacPermission) commonRun(profile string, testCases []TestCasePermission) {
+func (s *SuiteRbacPermission) commonRun(rbacProfile RBACProfile, testCases []TestCasePermission) {
 	t := s.T()
 
+	xrhidProfiles := []XRHIDProfile{XRHIDUser}
```
pvoborni, line 120: Is this line missing a service account (`xrhidProfiles := []XRHIDProfile{XRHIDUser, XRHIDServiceAccount}`) or is that added in the following PRs - I didn't read them yet?

(otherwise the new  `for` doesn't make sense)

pvoborni (commented):
Finished review. Added some comments and questions - mostly small stuff.

internal/test/smoke/rbac_permission_test.go
pvoborni: Answering to myself: yes it is added in next pr: https://github.com/podengo-project/idmsvc-backend/pull/233/commits/d7d1c3d6110e8f479fea77eb8e0cf02d0ab70b49#diff-c2ee44974ea374e676d4bfc79c174c3dc367af8da4b273e2f4a564528030e61d

internal/test/smoke/suite_base_test.go
```diff
@@ -37,6 +37,30 @@ import (
...
+	// XRHIDNothing does not make any action, letting a progressive change
+	XRHIDNothing XRHIDProfile = ""
+	// XRHIDNone no XRHID header is added
+	XRHIDNone XRHIDProfile = "None"
+	// XRHIDUser add a user XRHID header
+	XRHIDUser XRHIDProfile = "User"
+	// XRHIDServiceAccount add a service account XRHID header
+	XRHIDServiceAccount XRHIDProfile = "ServiceAccount"
+	// XRHIDSystem add a system XRHID header
+	XRHIDSystem XRHIDProfile = "System"
```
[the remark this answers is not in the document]
avisiedo: :+1: I have re-written in a different way; and you are right, basically it is an enumeration of supported xrhid profiles; **but** two values have a special meaning; the empty string fit when we don't initialize the value in a structure, the "None" has another special meaning that assure to delete the xrhid. They are chosen on purpose, to avoid the empty string because a forgotten and evoke to delete the xrhid header without that intention, and to delete specifically with the second value when required.

internal/test/smoke/suite_base_test.go
```diff
@@ -37,6 +37,30 @@ import (
...
+)
+
+type RBACProfile string
+
+const (
+	RBACSuperAdmin RBACProfile = mock_rbac.ProfileSuperAdmin
+	RBACAdmin      RBACProfile = mock_rbac.ProfileSuperAdmin
```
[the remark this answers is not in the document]
avisiedo: :+1: very good catch! many thanks!

internal/test/builder/api/builder_service_account_xrhid.go
```diff
@@ -0,0 +1,94 @@
...
+	"strconv"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/podengo-project/idmsvc-backend/internal/test/builder/helper"
+	builder_helper "github.com/podengo-project/idmsvc-backend/internal/test/builder/helper"
+	identity "github.com/redhatinsights/platform-go-middlewares/v2/identity"
+)
+
+type ServiceAcccountXRHID interface {
```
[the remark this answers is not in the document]
avisiedo: good catch! thanks mate!

internal/test/builder/api/builder_service_account_xrhid.go
```diff
@@ -0,0 +1,94 @@
...
+	WithAccountNumber(value string) ServiceAccountXRHID
+	WithEmployeeAccountNumber(value string) ServiceAccountXRHID
+	WithOrgID(value string) ServiceAccountXRHID
+	WithAuthType(value string) ServiceAccountXRHID
+
+	WithClientID(value string) ServiceAccountXRHID
+	WithUsername(value string) ServiceAccountXRHID
+}
+
+type serviceAcccountXRHID identity.XRHID
```
pvoborni, line 25: The three `c` are also here.

pvoborni (approved):
I'm approving as the minor naming issue is in internal type and can be fixed separately.

PR #234, opened 2024-06-15
bump(HMS-3968): github.com/redhatinsights/platform-go-middlewares/v2 v2.0.0-beta.2

avisiedo:
This change bump the platform-go-middlewares to `v2` which provide support for service accounts.

As no backward compatibility, we need to add some checks to avoid SIGVSEG situations and update unit tests to align with the new library and fix.


frasertweedale:
@avisiedo we have some conficts with the recent dependabot merges.  Hopefully a simple rebase!

avisiedo:
thanks @frasertweedale having a look! :+1:

internal/infrastructure/middleware/enforce_identity_test.go
```diff
@@ -482,64 +506,76 @@ func TestEnforceIdentityOrder(t *testing.T) {
 	}
 }
 
+// IdentityAlwaysTrue is a predicate that always return nil
+// so everything was ok.
+// data is the reference to the identity.Identity data.
+// Return nil on success or an error with additional
+// information about the predicate failure.
+func identityAlwaysTrue(*identity.XRHID) error {
```
pvoborni, line 514: The comment doesn't reflect the implementation. E.g. text `identityAlwaysTrue is a predicate that always return nil as if everything was ok` would be enough.

internal/infrastructure/middleware/enforce_identity_test.go
```diff
@@ -482,64 +506,76 @@ func TestEnforceIdentityOrder(t *testing.T) {
...
+	return nil
+}
+
+// NewIdentityAlwaysFalse is a predicate that always return
+// an error. It is useful for testing.
+func newIdentityAlwaysFalse(err error) func(*identity.XRHID) error {
```
pvoborni, line 520: The comment is bit incorrect as the function itself is not a predicate. I'd use e.g. `newIdentityAlwaysFalse returns a predicate that always return specified error`

Also, I'd prefer to specify the the return value of the function as `IdentityPredicate`

internal/infrastructure/middleware/enforce_identity_test.go
```diff
@@ -482,64 +506,76 @@ func TestEnforceIdentityOrder(t *testing.T) {
...
+	return func(*identity.XRHID) error {
+		return err
+	}
+}
+
 func TestNewEnforceOr(t *testing.T) {
+	type TestCaseGiven struct {
+		First  func(*identity.XRHID) error
+		Second func(*identity.XRHID) error
```
pvoborni, line 529: What about rather using:
```go
	type TestCaseGiven struct {
		First  IdentityPredicate
		Second IdentityPredicate
	}
```
IMHO is more readable.

internal/infrastructure/middleware/enforce_identity_test.go
```diff
@@ -482,64 +506,76 @@ func TestEnforceIdentityOrder(t *testing.T) {
...
 	type TestCase struct {
 		Name     string
-		Given    *identity.XRHID
+		Given    TestCaseGiven
 		Expected error
 	}
+	wrongSystemXRHID := api.NewUserXRHID().Build()
+	wrongSystemXRHID.Identity.Type = "System"
+	wrongUserXRHID := api.NewSystemXRHID().Build()
+	wrongUserXRHID.Identity.Type = "User"
```
pvoborni, line 539: Both `wrongSystemXRHID` and `wrongUserXRHID` are unused.

pvoborni (commented):
The dependency bump looks good. 

But the enforce_identity and test related changes seems almost belonging to a different PR. Commented on some smaller issues there.

pvoborni (approved):
LGTM

PR #233, opened 2024-06-15
test(HMS-3968): cover service accounts in smoke

avisiedo:
This change add service accounts to the rbac smoke
tests, and complete the existing smoke tests to
provide coverage for the service accounts where
required.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/232

PR #236, opened 2024-06-15
chore(HMS-3968): update test scripts

avisiedo:
This change add service account support to the helper scripts
at `./scripts/test/*.sh` files.

We can specify `XRHID_AS={user,system,service-account}`, by
default it is used `user` account.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/233

### DIFF
--- a/test/scripts/common.inc
+++ b/test/scripts/common.inc
@@ -24,6 +24,10 @@ BASEDIR="$(dirname "$(dirname "${SRCDIR}")")"
 REPOBASEDIR="$(git rev-parse --show-toplevel)"
 export REPOBASEDIR
 export XRHIDGEN="${REPOBASEDIR}/tools/bin/xrhidgen"
+# "jwt-auth" "cert-auth"
+export AUTH_TYPE="${AUTH_TYPE:-jwt-auth}"
+
+XRHID_CLIENT_ID="${XRHID_CLIENT_ID:-$(uuidgen)}"
 
 if [[ ! -x "${XRHIDGEN}" ]]; then
     error "${XRHIDGEN} is missing, run 'make install-tools'"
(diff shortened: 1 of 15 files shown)


test/scripts/common.inc
```diff
@@ -45,3 +47,30 @@ identity_user() {
 identity_system() {
     "${XRHIDGEN}" -org-id "${ORG_ID}" system -cn "6f324116-b3d2-11ed-8a37-482ae3863d30" -cert-type system | base64nowrap
 }
+
+identity_service_account() {
+    "${XRHIDGEN}" -org-id "${ORG_ID}" service-account -client-id "${XRHID_CLIENT_ID}" -username test | base64nowrap
```
pvoborni, line 52: The command should be:
```
"${XRHIDGEN}" -org-id "${ORG_ID}" -auth-type jwt-auth service-account -client-id "${XRHID_CLIENT_ID}" -username test | base64nowrap
```
I.e the `-auth-type jwt-auth` was added.

The reason is that when I tried e.g. `XRHID_AS=service-account ./test/scripts/local-domains-list.sh` I got 401 issue.  

Logs did not help me much as there was only error from user predicate and not the service account predicate.

So after small modification:
```go
func NewEnforceOr(predicates ...IdentityPredicate) IdentityPredicate {
	return func(data *identity.XRHID) error {
		var allErrs error
		for i := range predicates {
			if err := predicates[i](data); err == nil {
				return nil
			} else {
				allErrs = errors.Join(allErrs, err)
			}
		}
		return allErrs
	}
}
```

I now see issue:
```
level=ERROR msg="'user-and-sa-identity' IdentityPredicate failed: 'Identity.Type=ServiceAccount' is not 'User'\n'Identity.AuthType' is not 'jwt-auth'"
```

A Q can be also whether we must limit it to `jwt-auth`.

pvoborni (changes requested):
I didn't test all the scripts but the change is the same in each so testing few seems good enough. But it needs a change, see the comment.

test/scripts/common.inc
avisiedo: thanks for this; I like the idea to join the errors and provide full error information. Adding that. Considering the `jwt-auth` I have realized I missed some changes in the scripts (adding the suggested to specify  `AUTH_TYPE`).

IMHO we would need to limit to `jwt-auth` (user and service-account) and `cert-auth`; Adding the change too.

pvoborni: Might be good to do the additional changes in separate PR. Here we can just fix the `identity_service_account` and merge.

avisiedo:
I have added a few changes that I have realized when looking at the changes! thanks for the enhancements and for the review! :+1:

test/scripts/common.inc
avisiedo: I have added additional changes at: https://github.com/podengo-project/idmsvc-backend/pull/249 but I had some situation with different golang versions (1.21 and 1.22) when running the unit tests, that evoked different results for the echo router unit-tests.

pvoborni (approved):
LGTM

PR #249, opened 2024-06-29
fix(HMS-3968): multi-error enforce and minor fixes

avisiedo:
This change are a set of small changes observed during the
review at https://github.com/podengo-project/idmsvc-backend/pull/236

- Add multi-error for OR logical enforcement.
- Add check for Identity.AuthType==`cert-auth` enforcement for System endpoints.
- A unit test for the router started to fails. The fix is added here, but this was not observed on the pipeline.

### DIFF
--- a/internal/infrastructure/middleware/enforce_identity.go
+++ b/internal/infrastructure/middleware/enforce_identity.go
@@ -75,6 +75,9 @@ func EnforceSystemPredicate(data *identity.XRHID) error {
 	if data.Identity.Type != "System" {
 		return fmt.Errorf("'Identity.Type' must be 'System'")
 	}
+	if data.Identity.AuthType != "cert-auth" {
+		return fmt.Errorf("'Identity.AuthType' is not 'cert-auth'")
+	}
 	if data.Identity.System == nil {
 		return fmt.Errorf("'Identity.System' is nil")
 	}
--- a/internal/infrastructure/middleware/enforce_identity.go
+++ b/internal/infrastructure/middleware/enforce_identity.go
@@ -3,6 +3,7 @@ package middleware
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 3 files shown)


avisiedo:
I have observed different behavior for the unit-test for the public router, when using g1.21 and go1.22; I would like to make it run correctly in both versions. Currently draft state.

I want to use https://pkg.go.dev/runtime#Version as workaround until base container image is on go1.22.

internal/infrastructure/middleware/metrics_test.go
```diff
@@ -93,4 +93,10 @@ func TestMetricsMiddlewareWithConfigCreation(t *testing.T) {
 
 	assert.Equal(t, http.StatusOK, resp.Code)
 	assert.Equal(t, "Ok", resp.Body.String())
+
+	// Check skipper
+	path = "/ping"
+	resp = httptest.NewRecorder()
+	req = httptest.NewRequest(http.MethodGet, path, nil)
+	e.ServeHTTP(resp, req)
```
pvoborni, line 101: Is the test complete? I'm not sure how it is "checking the skipper".

internal/infrastructure/router/public_test.go
```diff
@@ -152,13 +153,24 @@ func TestNewGroupPublic(t *testing.T) {
 
 		// This routes are added when the group is created
 		appPrefix + appName + versionFull + "/*": {
-			"echo_route_not_found": "github.com/labstack/echo/v4.glob..func1",
+			"echo_route_not_found": "github.com/labstack/echo/v4.init.func1",
```
pvoborni, line 156: What is the reason for this test to be testing the handlers? I don't see the benefit.  From my PoV it is looking at too much details and it is not really testing any behavior.  It would be good enough to check that `e.Routes` contains all the expected routes with correct names and http methods. 

That way `TestNewGroupPublic` test would be simpler and would not rely on internal implementation of echo framework.

pvoborni (commented):
Providing some preliminary review.

internal/infrastructure/middleware/metrics_test.go
avisiedo: thanks! I have updated the test to include assertion for the Skipper call.

internal/infrastructure/router/public_test.go
avisiedo: thanks! I have updated the code to enhance the unit test and get rid of that specific details.

avisiedo:
The additional commits I will gather in a different PR; thanks for the review and advices :+1:

pvoborni (approved):
LGTM

PR #251, opened 2024-07-01
fix(HMS-2991): fail on changing of realm in domains PUT

pvoborni:
Return 400 bad request when realm name is about to be changed as it is not possible to change a realm of an IPA server.

Note: the ticket discusses implementation in interactor and using `readOnly` property in API spec.  But this this is no-go as the check needs to be performed after loading the entry from DB (thus after interactor and all middlewares). 

As for review of other fields, the fields are mostly limited by the data structure of requests. To me, only the the RealmName seems as the field to check. PATH method (`UpdateDomainUserRequest` struct` is already quite limited and all fields are changeable).

### DIFF
--- a/internal/handler/impl/domain_handler.go
+++ b/internal/handler/impl/domain_handler.go
@@ -330,6 +330,15 @@ func (a *application) UpdateDomainAgent(ctx echo.Context, domain_id uuid.UUID, p
 		)
 	}
 
+	if data.IpaDomain != nil && currentData.IpaDomain != nil &&
+		data.IpaDomain.RealmName != nil && currentData.IpaDomain.RealmName != nil &&
+		*data.IpaDomain.RealmName != *currentData.IpaDomain.RealmName {
+		return internal_errors.NewHTTPErrorF(
+			http.StatusBadRequest,
+			"'realm_name' may not be changed",
+		)
+	}
+
 	if err = a.fillDomain(currentData, data); err != nil {
 		return err
 	}
--- a/internal/test/smoke/suite_base_test.go
+++ b/internal/test/smoke/suite_base_test.go
@@ -855,3 +855,36 @@ func TestSuite(t *testing.T) {
 	suite.Run(t, new(SuiteRbacPermission))
 	suite.Run(t, new(SuiteSystemEndpoints))
 }
+
+// ErrorResponse represents the error response from the API
+//
+// Note: the structure is different than the public.ErrorResponse
+// as the API doesn't actually return a public.ErrorResponse which it should.
+type ErrorResponse struct {
+	Error   string `json:"error"`
+	Message string `json:"message"`
+}
+
+type BodyFuncErrorResponse func(t *testing.T, body *ErrorResponse) error
+
+func WrapBodyFuncErrorResponse(predicate BodyFuncErrorResponse) BodyFunc {
+	if predicate == nil {
+		return func(t *testing.T, body []byte) bool {
+			return len(body) == 0
+		}
+	}
+	return func(t *testing.T, body []byte) bool {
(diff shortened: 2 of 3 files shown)


internal/test/smoke/suite_base_test.go
```diff
@@ -855,3 +855,36 @@ func TestSuite(t *testing.T) {
 	suite.Run(t, new(SuiteRbacPermission))
 	suite.Run(t, new(SuiteSystemEndpoints))
 }
+
+// ErrorResponse represents the error response from the API
+//
+// Note: the structure is different than the public.ErrorResponse
+// as the API doesn't actually return a public.ErrorResponse which it should.
```
avisiedo, line 862: Right! Related to https://issues.redhat.com/browse/HMS-1668

internal/handler/impl/domain_handler.go
```diff
@@ -330,6 +330,15 @@ func (a *application) UpdateDomainAgent(ctx echo.Context, domain_id uuid.UUID, p
...
 	}
 
+	if data.IpaDomain != nil && currentData.IpaDomain != nil &&
+		data.IpaDomain.RealmName != nil && currentData.IpaDomain.RealmName != nil &&
+		*data.IpaDomain.RealmName != *currentData.IpaDomain.RealmName {
+		return internal_errors.NewHTTPErrorF(
+			http.StatusBadRequest,
+			"'realm_name' may not be changed",
+		)
+	}
```
avisiedo, line 340: optional: I suggest to extract in a function to reduce ciclomatic complexity indicator for the method. BTW the code is working as it is.

```suggestion
       // Check the realm does not change
       if err = a.updateDomainAgentCheckRealmName(data, currentData); err != nil {
               return err
       }
```

And the function just encapsulate the `if` statement at `domain_handler_private.go` file:

```go
func (a *application) updateDomainAgentCheckRealmName(data, currentData *model.Domain) error {
       if data.IpaDomain != nil && currentData.IpaDomain != nil &&
               data.IpaDomain.RealmName != nil && currentData.IpaDomain.RealmName != nil &&
               *data.IpaDomain.RealmName != *currentData.IpaDomain.RealmName {
               return internal_errors.NewHTTPErrorF(
                       http.StatusBadRequest,
                       "'realm_name' may not be changed",
               )
       }
       return nil
}
```

avisiedo (approved):
LGTM

Thanks for the change!

It works as it is, only one optional suggestion, but it works as it is.

----

Tested by:

```sh
$ ./test/scripts/local-domains-token.sh
$ TOKEN="...."  # The returned token
$ ./test/scripts/local-domains-register.sh "$TOKEN"
$ UUID="..." # The returned domain_id
$ ./test/scripts/local-domains-update.sh "$UUID"
$ # Edit and change realm field
$ vi test/data/http/update-rhel-idm-domain.json
$ # Run update operation
$ ./test/scripts/local-domains-update.sh "$UUID"
HTTP/1.1 400 Bad Request
Content-Type: application/json
X-Frame-Options: SAMEORIGIN
X-Rh-Insights-Request-Id: test_76bba2f175eb48f8af3b
Date: Thu, 04 Jul 2024 12:39:30 GMT
Content-Length: 46

{"message":"'realm_name' may not be changed"}
```

internal/handler/impl/domain_handler.go
pvoborni: I'll skip it as, too keep consistency, it would need to be done also in the block above that checks the `title`.  

+ it is end of day before my PTO so we don't have time for another review round. 

btw, if I'd do it, I'd name the method e.g., `realmNameNotChangedCheck` or `realmNameNotChangedGuard` - to be more self-explanatory.

PR #252, opened 2024-07-01
feat(HMS-3596): respect defined values in registration

pvoborni:
If a system makes the registration call (domains POST) with defined:

- title
- description
- auto_enrollment_enabled

Then the values are respected (unchanged).

If values are not defined or if title is empty then defaults are set.

### DIFF
--- a/internal/test/smoke/register_test.go
+++ b/internal/test/smoke/register_test.go
@@ -123,10 +123,10 @@ func (s *SuiteRegisterDomain) TestRegisterDomain() {
 					}
 
 					require.NotNil(t, body.Description)
-					assert.Equal(t, "", *body.Description)
+					assert.Equal(t, *bodyRequest.Description, *body.Description)
 
 					require.NotNil(t, body.AutoEnrollmentEnabled)
-					assert.False(t, *body.AutoEnrollmentEnabled)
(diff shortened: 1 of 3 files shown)


pvoborni:
Topic for discussion @avisiedo , @frasertweedale  . This patch is changing the behavior for `auto_enrollment_enabled` on top of the request in HMS-3596.  Current expectation is that after `POST /domains` it will be always be disabled and now it will follow what is provided or will set `false` if nothing is provided.

I think this might be beneficial for cases where PPL are setting up the systems programmatically (outside of wizard) as it might (in future with changes in `ipa-hcc register`) allow registration in 2 steps instead of 3. E.g.

* As Service-Account POST domains/token
* on IPA server: `ipa-hcc register $TOKEN --enable --title=$TITLE --description=$DESCRIPTION`

I.e. the 3 options are basically what is done in the wizard. And on CLI admins would need to do a PATCH request as SA or user (we can trust skipping this as they use $TOKEN obtained with similar rights).

frasertweedale:
I'm OK with this change - ipa-hcc leaves the field unset so it should default to False.  This ensures the required user experience for the standard workflow.

I will review and test the code changes shortly.

frasertweedale (approved):

PR #254, opened 2024-07-02
fix: request type of domains update PUT and PATCH

pvoborni:
This is a fix/refactoring where the `domain_handler` is changed to use correct request type instead of `public.Domain` for PATCH and PUT. 

The `domains_interactor` is modified to reflect the changes.

This was made when working on HMS-2991 but it turned out that these changes are not needed for the ticket.


frasertweedale (approved):
Diff makes sense, new names are a bit easier to comprehend.

PR #255, opened 2024-07-03
fix(HMS-4373): small changes

avisiedo:
This change include:

- Relax public router test so the goal of checking the routes is achieved, but we don't enter on to deep framework details.
- Fix a go vet violation at the metrics middleware where one of the implicit call is not deferred.
- Simplify Skipper invocation for the metrics middleware and add unit test for it.


pvoborni (approved):
LGTM

PR #262, opened 2024-07-15
refactor: use log/slog

avisiedo:
Use the `log/slog` version to keep the
consistency cross the repository. Use different
packages could evoke the situation described at:

https://github.com/golang/go/issues/64015


aadhikar (approved):
LGTM!

PR #266, opened 2024-07-16
Update IQE_PLUGINS param in TESTING.md

aadhikar:


avisiedo (approved):
LGTM

Thanks for the change; some parts of the documents could need update, so the more it is updated the better.

I appreciate it!

PR #270, opened 2024-07-17
bump(rhin-1706): base container image

avisiedo:
This change bump the base container image to the
current most recent version.


aadhikar (approved):
LGTM!

PR #271, opened 2024-07-17
bump(rhin-1706): python requirements

avisiedo:
This change bump the dependencies to the latest versions but 'pre-commit' which has a bigger version change.


aadhikar (approved):
LGTM!

PR #272, opened 2024-07-17
HMS-4436 fix: IQE test failing in Stage-Testing SaaS pipeline

aadhikar:
tests/test_domain_basic.py::test_domain_register[ViaWebUI] is failing with ERROR Exception looking up a domain within
context: <iqe.base.application.implementations.web_ui.ViaWebUI

Element is present but the click event is not working.


avisiedo (approved):
LGTM

Thanks for the change!

PR #278, opened 2024-07-29
Revert back to old image as the latest image is not stable

aadhikar:
The pipeline is paused at INFO: sending shutdown request to selenium... 
The issue was with tests, not images, so reverting to the old image.


pvoborni (approved):
lGTM

PR #283, opened 2024-08-02
fix: url to renew token

avisiedo:
It was detected a wrong url into the
scripts/mk/private.example.mk file which
is used to get started the configuration
for the development environment.

This change update the url to renew the
token when we could need so it is easy
to make the operation for the developer.


pvoborni (approved):
LGMT

PR #284, opened 2024-08-05
fix(HMS-4522): use contextual log in middlewares

avisiedo:
If a middleware is configured after the contextual log, then it should use it, so the traceability
information is printed out with the messages.

This change add the above for enforcement and rbac middleware.

https://issues.redhat.com/browse/HMS-4522


pvoborni (approved):
Works fine. LGTM.

PR #288, opened 2024-08-06
feat(HMS-4520): aggregate identity information

avisiedo:
This change aggregate the org_id and the identity
information to the contextual log, so the information
is printed out for any log message.

### DIFF
--- a/internal/infrastructure/middleware/enforce_identity.go
+++ b/internal/infrastructure/middleware/enforce_identity.go
@@ -5,12 +5,24 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/labstack/echo/v4"
 	echo_middleware "github.com/labstack/echo/v4/middleware"
 	"github.com/podengo-project/idmsvc-backend/internal/api/header"
 	app_context "github.com/podengo-project/idmsvc-backend/internal/infrastructure/context"
 	identity "github.com/redhatinsights/platform-go-middlewares/v2/identity"
+)
+
+// https://github.com/RedHatInsights/identity-schemas/blob/main/3scale/schema.json#L13
+// https://github.com/RedHatInsights/identity-schemas/blob/main/turnpike/schema.json#L7
+const (
+	identityTypeUser           = "User"
+	identityTypeServiceAccount = "ServiceAccount"
+	identityTypeSystem         = "System"
+	identityTypeX509           = "X509"
+	identityTypeAssociate      = "Associate"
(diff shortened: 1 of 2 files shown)


internal/infrastructure/middleware/enforce_identity.go
```diff
@@ -13,6 +14,15 @@ import (
 	identity "github.com/redhatinsights/platform-go-middlewares/v2/identity"
 )
 
+// https://github.com/RedHatInsights/identity-schemas/blob/main/3scale/schema.json#L21
+const (
+	identityTypeUser          = "User"
+	identityTypeServieAccount = "ServiceAccount"
```
pvoborni, line 20: Typo: `identityTypeServieAccount` vs `identityTypeServiceAccount`

internal/infrastructure/middleware/enforce_identity.go
```diff
@@ -13,6 +14,15 @@ import (
 	identity "github.com/redhatinsights/platform-go-middlewares/v2/identity"
 )
 
+// https://github.com/RedHatInsights/identity-schemas/blob/main/3scale/schema.json#L21
```
pvoborni, line 17: Line 21 points to `auth_type` and not `type`. Also `type` there has only 3 values - `User`, `ServiceAccount`, `System`. 

The: `X509` and `Associate` are bit weird. It's coming from  in https://github.com/RedHatInsights/platform-go-middlewares/blob/master/identity/identity.go#L101 right?  I also saw them used as type in https://github.com/RedHatInsights/platform-go-middlewares/blob/master/identity/identity_suite_test.go so I guess we should rather work with them - so OK.

pvoborni (commented):
Mostly looks good to me, didn't test yet.

internal/infrastructure/middleware/enforce_identity.go
avisiedo: thanks! Updated!

internal/infrastructure/middleware/enforce_identity.go
avisiedo: good catch! I have updated the link! thanks!

avisiedo: you are right; `X509` and `Associate` are comming from the link you referenced; I added them as I have seen in the json schema definition at: https://github.com/RedHatInsights/identity-schemas/blob/main/turnpike/schema.json just for considering all the potential values and reflect the mapping in the log, and avoid `unknown` for them.

pvoborni (approved):
The PR LGTM, it works.  

How do we restart the CI job?

PR #289, opened 2024-08-06
chore(HMS-4521): clowder - default logging level set to info

pvoborni:
It was agreed as the least-verbose level accross all envs.


avisiedo (approved):
LGTM

PR #290, opened 2024-08-07
fix(HMS-4544): remove RequestID middleware

avisiedo:
This change remove the RequestID middleware as
it would be duplicated then in the response to
then end user.

Credits on @pvoborni


pvoborni (approved):
LGTM

PR #291, opened 2024-08-07
feat(HMS-4545): add source ip to logs

avisiedo:
This change add the source_ip aggregated field to the logs.

https://issues.redhat.com/browse/HMS-4545

### DIFF
--- a/internal/infrastructure/middleware/logger.go
+++ b/internal/infrastructure/middleware/logger.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"log/slog"
+	"strings"
 
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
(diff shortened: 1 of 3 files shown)


internal/infrastructure/middleware/logger.go
```diff
@@ -31,11 +41,13 @@ func ContextLogConfig(cfg *LogConfig) echo.MiddlewareFunc {
 			}
 
 			requestID := c.Request().Header.Get(header.HeaderXRequestID)
+			source := getSource(c.Request().Header.Get(header.HeaderXForwardedFor))
 
 			// Let print the request-id for every call of the
 			// request log
 			logger := slog.Default().With(
 				slog.String("request_id", requestID),
+				slog.String("source", source),
```
pvoborni, line 50: Atm, `source` is used for the location in source code of the log call, so we need a different key to not conflict.

avisiedo: I used initially `source_ip` but after observe the examples at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For I step back on that name.

Proposal: what about `source_host`??

pvoborni: both `source_ip`  and `source_host` sounds good to me. I'd prefer `source_ip` if the content is an IP address.  If it can contain also hostname then `source_host` might be better.

In the examples I saw only IP addresses. Was there any mention that it might be something different?

pvoborni (approved):
LGTM

PR #292, opened 2024-08-08
feat(HMS-4532): pendo client

avisiedo:
This change implement a pendo client,
and add the necessary changes to
update the config template, and read
the pendo configuration.

This add only the used endpoint.

https://issues.redhat.com/browse/HMS-4532

PR #294, opened 2024-08-09
feat(HMS-4532): mock pendo

avisiedo:
Quick mock service to mock pendo, to prepare
the repository to use it on smoke tests.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/292

https://issues.redhat.com/browse/HMS-4532

### DIFF
--- a/cmd/mock-pendo/main.go
+++ b/cmd/mock-pendo/main.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/podengo-project/idmsvc-backend/internal/config"
+	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/logger"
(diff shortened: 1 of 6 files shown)


pvoborni:
I don't think that we need mock like this atm as mocking the client or using dump might be good enough atm. But if we would like to use a mock service, why are we creating custom mock services and not using https://github.com/mock-server/mockserver that is already partly configured in our compose https://github.com/podengo-project/idmsvc-backend/blob/main/deployments/docker-compose.yaml#L7 . I think adding configuration to it might make the trick.

avisiedo:
About mock-server; it is true it is partially configured. What I recall when I added, I needed more knowledge for using it; I was capable to set up a mock response with it, but I had some situation related with preparing additional responses and with the sequence. I would need to repeat and try to set-up the process with mock-server.

frasertweedale:
Should we close this PR?  We can resurrect it later if we decide to resume the mock effort, using the same approach.

avisiedo:
:+1: closed

PR #295, opened 2024-08-09
feat(HMS-4532): integrate pendo client

avisiedo:
This change instantiate the client pendo
component and inject to make it available at
the application service scope.

More changes could be necessary depending on
where it will be used.


pvoborni (commented):
I look at the changes done in the last 2 commits and they look good to me. Also tested in local, that client starts correctly even when Pendo config values are not set. 

I'll wait for #292 to be merged, this rebased and then I'll approve (unless some other issue appears).

pvoborni (approved):
LGTM

PR #298, opened 2024-08-16
feat(HMS-4503): add Pendo track event for host-conf endpoint

pvoborni:
For measuring engagement metrics.
    
The `sendPendoTrackEvent` function should be reusable in different handlers.

The PR contains a couple prerequisites:
- move of GetPrincipal to be usable outside of the middleware
- generated mock Pendo client
- log of track event name in `PendoClientDump`


pvoborni:
It's also possible to implement in a middleware. There the advantage could be that we might be able to send TrackEvent when issue in some other middleware happens(e.g. 404 bad request with request not fulfilling schema).

internal/handler/impl/host_handler.go
```diff
@@ -13,11 +13,31 @@ import (
 	"gorm.io/gorm"
 )
 
+const (
+	pendoHostConfSuccess = "HostConf"
```
pvoborni, line 17: I'm thinking about name spacing the track event. e.g. `idmsvc-host-conf` or `idmsvc-HostConf`, wdyt?

avisiedo, line 17: Wondering if make sense to set the value to `HostConfSuccess` to be aligned with the next const value; I think it could provide more semantic; what do you think?

internal/handler/impl/host_handler.go
```diff
@@ -13,11 +13,31 @@ import (
...
 	ctx echo.Context,
 	inventoryId public.HostId,
 	fqdn string,
 	params public.HostConfParams,
+) error {
+	err := a.hostConf(ctx, inventoryId, fqdn, params)
+	if err != nil {
+		sendPendoTrackEvent(ctx, a.pendo, pendoHostConfFailed)
+		return err
+	}
```
avisiedo, line 31: nit: suggestion to express in the form below:

```suggestion
	if err := a.hostConf(ctx, inventoryId, fqdn, params); err != nil {
		sendPendoTrackEvent(ctx, a.pendo, pendoHostConfFailed)
		return err
	}
```

internal/handler/impl/utils.go
```diff
@@ -20,3 +24,27 @@ func getXRHID(ctx echo.Context) (*identity.XRHID, error) {
...
+		Event:     event,
+		VisitorID: header.GetPrincipal(xrhid),
+		Timestamp: time.Now().UTC().Unix(),
+	}
+
+	err = client.SendTrackEvent(c, &track)
+	if err != nil {
+		logger.Warn(err.Error())
+	}
+}
```
avisiedo, line 50: idea: what if we add the function as a private method, the `client` is available as part of the attributes of the instance, and we would have it available:

for instance:

```suggestion
func (a *application) sendPendoTrackEvent(ctx echo.Context, event string) error {
	var (
		xrhid *identity.XRHID
		err   error
	)
	c := ctx.Request().Context()
	logger := app_context.LogFromCtx(c)

	if xrhid, err = getXRHID(ctx); err != nil {
		logger.Error(err.Error())
		return err
	}

	track := pendo.TrackRequest{
		AccountID: xrhid.Identity.OrgID,
		Type:      pendoTypeTrack,
		Event:     event,
		VisitorID: header.GetPrincipal(xrhid),
		Timestamp: time.Now().UTC().Unix(),
	}

	if err = a.pendo.SendTrackEvent(c, &track); err != nil {
		logger.Warn(err.Error())
		return err
	}

	logger.Debug("sendPendoTrackEvent was success",
		slog.String("Type", track.Type),
		slog.String("Event", track.Event))
	return nil
}
```

In the above case I would keep in a different file, for instance `internal_methods.go` (not using `private.go` as already we have `private_handler.go`, maybe I didn't choose the best name when the `private_handler.go` file was added).

internal/test/smoke/suite_base_test.go
```diff
@@ -135,8 +137,10 @@ func (s *SuiteBase) SetupTest() {
 		panic(err)
 	}
 	rbac := client_rbac.New(s.cfg.Clients.RbacBaseURL, rbacClient)
-	pendo := client_pendo.NewClient(s.cfg)
-	s.svc = service_impl.NewApplication(ctx, s.wg, s.cfg, s.db, inventory, rbac, pendo)
+	if s.PendoClient == nil {
+		s.PendoClient = client_pendo.NewClient(s.cfg)
+	}
```
avisiedo, line 142:
```suggestion
	s.PendoClient = mock_pendo.NewPendo(t)
```

I would suggest to use the mock generated by mockery tool, and this would make it available for any smoke test.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -19,6 +22,11 @@ type SuiteSystemEndpoints struct {
 	domain *public.Domain
 }
 
+func (s *SuiteSystemEndpoints) SetupTest() {
+	s.PendoClient = mock_pendo.NewPendo(s.T())
```
avisiedo, line 26: If the above change on the SuiteBase is added, this line wouldn't be necessary.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -87,20 +95,60 @@ func (s *SuiteSystemEndpoints) TestHostConfExecute() {
...
-	require.NoError(t, err)
-	require.NotNil(t, res)
-	require.Equal(t, http.StatusOK, res.StatusCode)
+
+	t.Run("HostConf success", func(t *testing.T) {
+		// Expect success HostConf event to be sent to Pendo
+		mockPendo, ok := s.PendoClient.(*mock_pendo.Pendo)
+		require.True(t, ok)
+		mockPendo.On("SendTrackEvent", mock.Anything, mock.MatchedBy(func(r *pendo.TrackRequest) bool {
+			return (r.Event == "HostConf" &&
```
avisiedo, line 106: I suggest to use `PendoHostConfSuccess` constant; that would keep the test in sync with the event value. I suggest to move to a common place the constant (thinking on `internal/interface/client/pendo` package, but I recognize not 100% happy with that suggested location as it would be providing service dependent values at a place which is client independent place).

internal/test/smoke/system_endpoints_test.go
```diff
@@ -87,20 +95,60 @@ func (s *SuiteSystemEndpoints) TestHostConfExecute() {
...
+
+		mockPendo.AssertExpectations(t)
+	})
+
+	t.Run("HostConf failure", func(t *testing.T) {
+		// Expect failure HostConf event to be sent to Pendo
+		mockPendo, ok := s.PendoClient.(*mock_pendo.Pendo)
+		require.True(t, ok)
+		mockPendo.On("SendTrackEvent", mock.Anything, mock.MatchedBy(func(r *pendo.TrackRequest) bool {
+			return (r.Event == "HostConfFailed" &&
```
avisiedo, line 132: Similar to the above; by moving the constants to a common place, we would use here too, and any change to the value would not impact the test.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -87,20 +95,60 @@ func (s *SuiteSystemEndpoints) TestHostConfExecute() {
...
-		"client."+s.domain.DomainName,
-		builder_api.NewHostConf().
-			WithDomainName(pointy.String(s.domain.DomainName)).
-			WithDomainType(&domainType).
-			Build())
-	require.NoError(t, err)
-	require.NotNil(t, res)
-	require.Equal(t, http.StatusOK, res.StatusCode)
+
+	t.Run("HostConf success", func(t *testing.T) {
```
avisiedo, line 101: I suggest to split this in two different test methods: `TestHostConfExecuteSuccess` and `TestHostConfExecuteFailure`, and move the common code to a private method. Only for consistency; I am not familiar using directly the `t.Run` method, but happy to know more about it.

Every `Test...` method would have a fresh and clean mock instance to add the expectations for SendTrackEvent.

internal/test/smoke/system_endpoints_test.go
```diff
@@ -87,20 +95,60 @@ func (s *SuiteSystemEndpoints) TestHostConfExecute() {
...
+		// Expect success HostConf event to be sent to Pendo
+		mockPendo, ok := s.PendoClient.(*mock_pendo.Pendo)
```
avisiedo, line 103: observation: I could be wrong, but maybe the type assertion is not necessary if the mock is declared as `*mock_pendo.Pendo` at `SuiteBase`.

avisiedo (changes requested):
:+1: Thanks for the changes.

I have learned about leverage `mock.MatchedBy`.

A few change suggestions.

internal/handler/impl/host_handler.go
avisiedo: Given the two options above, I'd rather `idmsvc-host-conf` to `idmsvc-HostConf`; rational: the second mix kebab case and pascal case.

* snake case: `number_of_donuts` or `NUMBER_OF_DONUTS`.
* kebab case: `number-of-donuts` or `NUMBER-OF-DONUTS`.
* camel case: `numberOfDonuts`.
* pascal case: `NumberOfDonuts`.

https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/

Looking at pendo, Track Events, I would like to know where would this value is mapped; if this value would be mapped to the `Name` column I can see a mix of different naming conventions; so I would say:
* I would ask to analytics mates if there are some naming convention to be aligned with it.
* If not naming convention defined or used and we are free to choose the one that fit better for us, I would say `kebab-case` as I have seen many events with the Name column using that convention; additionally, we could define scope as we progress (for instance, `idmsvc-host-conf-success` or `idmsvc-host-conf-failure`), and IMHO I would add a final label to indicate what event is happening: `success` or `failure`.

Based on the last example `idmsvc-hostconf-{success,failure}` the pattern would looks like: `<service>-<label-path-of-my-event>-<label-to-give-semantic-to-the-event>`; examples:
* `idmsvc-host-conf-success`, where:
  * `service` = `idmsvc`
  * `label-path-of-my-event` = `host-conf` or `auto-enroll-host-conf` or `ipa-auto-enroll-host-conf`
  * `label-to-give-semantic-to-the-event` = `success` or `failure`

internal/handler/impl/utils.go
```diff
@@ -20,3 +24,27 @@ func getXRHID(ctx echo.Context) (*identity.XRHID, error) {
...
+		return
+	}
+
+	track := pendo.TrackRequest{
+		AccountID: xrhid.Identity.OrgID,
+		Type:      "track",
+		Event:     event,
+		VisitorID: header.GetPrincipal(xrhid),
+		Timestamp: time.Now().UTC().Unix(),
+	}
```
avisiedo, line 44:
```suggestion
	track := pendo.TrackRequest{
		AccountID: xrhid.Identity.OrgID,
		Type:      pendoTypeTrack,
		Event:     event, // Core event column in Pendo?
		VisitorID: header.GetPrincipal(xrhid),
		Timestamp: time.Now().UTC().Unix(),
		URL:       ctx.Request().URL.String(), // not 100% sure if this would set the public or the in-cluster URL.
		UserAgent: ctx.Request().UserAgent(),  // not 100% sure, because the several reverse proxies
		IP:        GetRealIP(ctx), // Read real IP from headers
		// What additional properties to add if any?
		Properties: map[string]any{
			"my-property-1": true,
			"my-property-2": "some-value",
		},
		// What additional context to add if any?
		Context: map[string]any{
			"my-context-1": 5,
			"my-context-2": "another-value",
		},
		Title: "Insights IDMSVC Track Event",
	}
```

avisiedo: I committed the suggestion by mistake; I have rolled back the branch to the state before the commit, and re-opened this suggestion; sorry for the inconveniences.

internal/test/smoke/suite_base_test.go
pvoborni: Do we want to use the mock in all smoke tests?  The current implementation allows to choose. So if the mock is not se then it will use the `NewClientDump()` or real `NewClient()`.

internal/test/smoke/system_endpoints_test.go
pvoborni: If it would be `*mock_pendo.Pendo` then it would not be necessary, but the type is the interface and thus it is currently necessary.

internal/test/smoke/system_endpoints_test.go
pvoborni: Do we have some high level package where similar stuff can belong. E.g. `internal.domain.pendo` and we can have all track events on single place.

internal/handler/impl/utils.go
pvoborni: I don't see a big need to move it as method of app. 

If we would move it then it might be to have more specific filename. As both `utils` or  `internal`, "private" forms coincidental cohesion which is better to be avoided.   E.g. `pendo_helpers.go` might be better.

internal/handler/impl/host_handler.go
pvoborni: Fixed, used `idmsvc-host-conf-success` and `idmsvc-host-conf-failure`.

internal/handler/impl/host_handler.go
pvoborni: Fixed.

pvoborni:
All review comments are either addressed or commented on (or their related were commented on).

internal/test/smoke/suite_base_test.go
avisiedo: thanks, I didn't see in that way! I agree with the current change!

internal/test/smoke/system_endpoints_test.go
avisiedo: Based on the previous comment! it is ok as it is! thanks!

internal/usecase/client/pendo/pendo_client_dump.go
```diff
@@ -41,7 +42,7 @@ func (c *pendoClientDump) SetMetadata(ctx context.Context, kind pendo.Kind, grou
 // SendTrackEvent
 func (c *pendoClientDump) SendTrackEvent(ctx context.Context, track *pendo.TrackRequest) error {
 	logger := app_context.LogFromCtx(ctx)
-	logger.Debug("pendo SendTrackEvent called")
+	logger.Debug(fmt.Sprintf("pendo SendTrackEvent called: %s", track.Event))
```
avisiedo, line 45: optional: use structured primitives for slog.

```suggestion
	logger.Debug("pendo SendTrackEvent called",
		slog.String("Event", track.Event),
	)
```

avisiedo (approved):
Thanks for the clarification on some concerns.

I have added an optional change when using structured logs feel free to add it.

The PR as it is LGTM; thanks for the changes.

internal/test/smoke/system_endpoints_test.go
avisiedo: :+1: good idea! that is a nice place land that values and share them cross the application.

PR #306, opened 2024-08-22
HMS-4554 Added SE_NODE_SESSION_TIMEOUT Environment Variable for Enhan…

aadhikar:
…ced Stability in IdM Test Environment

Description: Introduced the SE_NODE_SESSION_TIMEOUT environment variable, setting its value to 600 seconds. This change addresses the issue encountered during testing with a newly installed IdM server, where the extended installation time caused the browser to enter an unknown state, leading to test failures. By increasing the session timeout, we ensure that tests remain stable and pass consistently, even with the added server installation duration.


pvoborni (approved):
LGTM

PR #307, opened 2024-08-26
fix: timeout duration of Pendo client

pvoborni:
The timeout was not using correct duration unit as default is Nanosecond.

So when used with defaults, the request almost immediately timed out.


frasertweedale (approved):

frasertweedale:
Good catch, thanks!

PR #313, opened 2024-08-30
feat(HMS-4531): inject pendo env vars

avisiedo:
The change add the necessary environment variables to the deployment
so the client component can instantiate correctly.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/311

https://issues.redhat.com/browse/HMS-4531

### DIFF
--- a/deployments/clowder.yaml
+++ b/deployments/clowder.yaml
@@ -102,6 +102,20 @@ objects:
                     name: app-secret
               - name: CLIENTS_RBAC_BASE_URL
                 value: "${CLIENTS_RBAC_BASE_URL}"
+              - name: CLIENTS_PENDO_BASE_URL
+                value: "${CLIENTS_PENDO_BASE_URL}"
+              - name: CLIENTS_PENDO_API_KEY
+                valueFrom:
+                  secretKeyRef:
+                    key: integration-key
+                    name: pendo-creds
+                    optional: true
+              - name: CLIENTS_PENDO_TRACK_EVENT_KEY
+                valueFrom:
+                  secretKeyRef:
+                    key: track-event-secret
+                    name: pendo-creds
+                    optional: true
             resources:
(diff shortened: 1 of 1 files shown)


deployments/clowder.yaml
```diff
@@ -154,6 +168,20 @@ objects:
                     name: app-secret
               - name: CLIENTS_RBAC_BASE_URL
                 value: "${CLIENTS_RBAC_BASE_URL}"
+              - name: CLIENTS_PENDO_BASE_URL
```
pvoborni, line 171: A Q. The `db-tool jwk refresh` command is not planned to use the direct send to Pendo. I.e. the config is not needed here.  But that was the same with RBAC. Or was it needed?

avisiedo: You are totally right! thanks! I will remove from the containers that does not need them!

deployments/clowder.yaml
```diff
@@ -102,6 +102,20 @@ objects:
...
+                value: "${CLIENTS_PENDO_BASE_URL}"
+              - name: CLIENTS_PENDO_INTEGRATION_KEY
+                valueFrom:
+                  secretKeyRef:
+                    key: integration-key
+                    name: pendo-creds
+                    optional: true
+              - name: CLIENTS_PENDO_TRACK_EVENT_SECRET
```
pvoborni, line 113: In PR #231 this env var ends with `KEY` and not `SECRET`.  And the other is already name: `CLIENTS_PENDO_API_KEY` but here it is `CLIENTS_PENDO_INTEGRATION_KEY`.

I.e. we have a mismatch and need to unite.

avisiedo: Right! Updated description to reference #311 and label! 

Updated!

pvoborni (approved):
LGTM

PR #315, opened 2024-09-04
Add SonarQube job to GitLab CI

frasertweedale:
Add SonarQube SAST job to git GitLab CI (needs to be GitLab because it reports to the
Red Hat internal SonarQube instance).

The GitLab `test` CI job has gone stale and doesn't work (due to no database service, and
other gaps).  Because we use GitHub for the main CI, I just disabled the tests in GitLab
rather than wrestling with them.

Because GitLab is mirroring from idmsvc, CI jobs do not run automatically.  I have set up
and **hourly** GitLab scheduled pipeline to run on the `main` branch.  After this changed
has been merged and everything is verified to be working, I will wind that back to daily
or weekly.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,6 +38,8 @@ check missed commits:
 
 run tests:
   stage: test
+  rules:
+    - when: never
   tags:
     - shared
   needs: []
@@ -58,6 +60,5 @@ compile project:
   stage: build
   tags:
     - shared
-  needs: ["run tests"]
   script:
     - make build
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -5,6 +5,8 @@ stages:
   - check
   - test
   - build
+  - sonarqube-check
+  - sonarqube-vulnerability-report
```
pvoborni, line 9: This stage looks as leftover.

frasertweedale:
> LGTM, I'm assuming it is the exact same code except the removed report as in GitLab.

Indeed.  I even left the reference to the `sq-test` branch in there :)

PR #316, opened 2024-09-05
feat(HMS-4617): tls in-cluster communication

avisiedo:
Add support for encrypted in-cluster communication.

TODO:
- [x] Update Dockerfile to env var for certificate paths.
- [x] Check in dev cluster.
- [x] Add advised changes and refactor to enhance code.
- [x] Check the routine to update `SSL_CERT_DIR`

https://issues.redhat.com/browse/HMS-4617

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,6 +9,7 @@ package config
 import (
 	"log/slog"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -47,6 +48,10 @@ const (
 	DefaultAcceptXRHFakeIdentity = false
 	// DefaultValidateAPI is true
 	DefaultValidateAPI = true
+
+	// EnvSSLCertDirectory environment variable that provides
+	// the paths for the CA certificates
+	EnvSSLCertDirectory = "SSL_CERT_DIR"
 )
 
 type Config struct {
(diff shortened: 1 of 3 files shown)


internal/config/config.go
```diff
@@ -310,6 +320,41 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 	// Metrics configuration
 	v.Set("metrics.path", clowderConfig.MetricsPath)
 	v.Set("metrics.port", clowderConfig.MetricsPort)
+
+	// Set TLS configuration
+	processPublicEndpoint("rbac", "v1", v, clowderConfig)
+}
+
+func processPublicEndpoint(serviceName, version string, v *viper.Viper, clowderConfig *clowder.AppConfig) {
```
pvoborni, line 328: From the method signature, I did not get that it sets the configuration value in `v` based on the service name. I.e. it is a bit cryptic and I needed to read it whole to understand it. The cause might be that the function is solving multiple problems. 

I'm thinking of approach:

```go

func getEndpointPath(serviceName, version String, clowderConfig *clowder.AppConfig) String {
   // most of the code here

   // here the question is whether is should also return a non-TLS path if the service doesn't have TLS path
}

func updateServiceBasePath(serviceName, version string, v *viper.Viper, clowderConfig *clowder.AppConfig) {
    basePath := getEndpointPath(serviceName, version, v, clowderConfig)
    if basePath != "" {
        // Q is whether this should rather be a param to the updateServiceBasePath function
        // and we could skip the "calculation". 
        configPath := getServiceBaseURLConfigPath(serviceName, version)  
        slog.Debug("base url for service " + serviceName + "  updated to '" + basePath + "'")
		v.Set(configPath, basePath)
    }
}

```

internal/config/config.go
```diff
@@ -310,6 +320,41 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
...
+		panic("'v' is nil")
+	}
+	if clowderConfig == nil {
+		panic("'clowderConfig' is nil")
+	}
+	if clowderConfig.TlsCAPath != nil {
+		paramPath := "clients." + serviceName + "_base_url"
+		serviceEndpoint := getPublicEndpoint(serviceName, clowderConfig)
+		if serviceEndpoint != nil {
+			serviceBaseURLString := "https://" + serviceEndpoint.Hostname
```
pvoborni, line 345: The protocol should be `http` if TlsPort is not defined.   https://gitlab.cee.redhat.com/insights-platform/advisor-backend/-/merge_requests/490/diffs#652c5e5dd60867263fdb20ba5922d3d34a615885_102_103

pvoborni (commented):
With this patch it will do the following, right?:

- use value from `CLIENTS_RBAC_BASE_URL`
- override the value by the endpoint url if present. 

Do we want this?  In general, I think, we want to use `CLIENTS_RBAC_BASE_URL` for non-{prod,stage,ephemeral } envs. I.e. testing envs.  Or it might be used in prod/stage to override the endpoint we get in case of some issue (unlikely).

From this PoV, I'd stop using `CLIENTS_RBAC_BASE_URL` in stage/prod and have it only for podman-compose purposes with mock rbac. I.e. IMHO it should override the values from clowder.

internal/config/config.go
avisiedo: thanks! updated with the current state!

avisiedo:
> With this patch it will do the following, right?:
> 
> * use value from `CLIENTS_RBAC_BASE_URL`
> * override the value by the endpoint url if present.
> 
> Do we want this? In general, I think, we want to use `CLIENTS_RBAC_BASE_URL` for non-{prod,stage,ephemeral } envs. I.e. testing envs. Or it might be used in prod/stage to override the endpoint we get in case of some issue (unlikely).
> 
> From this PoV, I'd stop using `CLIENTS_RBAC_BASE_URL` in stage/prod and have it only for podman-compose purposes with mock rbac. I.e. IMHO it should override the values from clowder.

:+1: I removed the values from the `deployments/clowder.yaml` file, it still read from environment variable if it exists to allow local debugging and development.

* (viper)Load values from `configs/config.yaml`; I have seen this is copied in the container. `config.yaml:clients.rbac_base_url`.
* (viper)Overrides values from environment variables if a default value is defined in the viper configuration. `setDefaults`
* (our code)Overrides values with the configuration injected from clowder. `setClowderConfiguration` if `clowder.IsClowderEnabled()`.
  * If the endpoint is found in the clowder configuration, then override the final value with the endpoint information generating a base url from it.
  
I have checked locally by setting `ACG_CONFIG` environment variable pointing out to an example `cdappconfig.json` file. I can see that locally it seems to work, but I am trying to see the debug message in the dev cluster, indicating that the configuration value is override.

internal/config/config.go
```diff
@@ -310,6 +311,144 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
...
+		// NOTE
+		//
+		// ApiPath is deprecated, when removed from the operator we could remove this block. We will
+		// get an error compiling when that situation happens.
+		// This is added as fallback mechanism until completly removed the attribute.
+		//
+		// https://github.com/RedHatInsights/clowder/blob/master/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go#L152
+		return strings.TrimSuffix(serviceEndpoint.ApiPath, "/") +
+			"/" + version
+	}
```
avisiedo, line 436: from collab review: better remove the handling of the deprecated code.

deployments/clowder.yaml
```diff
@@ -235,7 +233,5 @@ parameters:
     value: "true"
     description: |
       It allows to enable / disable RBAC middleware.
-  - name: CLIENTS_RBAC_BASE_URL
-    required: true
-    description: |
-      Point out to the rbac service base url
+  - name: SSL_CERT_DIR
```
frasertweedale, line 236: Could you please add a comment explaining where the TLS CA certs are coming from (clowder operator populates the cdapp directory with trust roots).

build/package/Dockerfile
```diff
@@ -19,6 +19,7 @@ FROM registry.access.redhat.com/ubi9-minimal@sha256:a7d837b00520a32502ada85ae339
 LABEL idmsvc-backend=backend
 # https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant
 ENV OPENSSL_FORCE_FIPS_MODE=1
+ENV SSL_CERT_DIR="/etc/ssl/certs:/etc/pki/tls/certs:/system/etc/security/cacerts:/cdapp/certs"
```
frasertweedale, line 22: Could you please add a comment explaining where the TLS CA certs are coming from (clowder operator populates the cdapp directory with trust roots).

internal/config/config.go
```diff
@@ -310,6 +311,144 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
...
+			slog.String("TlsCAPath", *clowderConfig.TlsCAPath),
+		)
+	}
+
+	// Override client base url configuration from clowder when available
+	updateServiceBasePath("rbac", "v1", v, clowderConfig)
+}
+
+// guardUpdateServiceBasePath raise a panic when some of the arguments for
+// processPublicEndpoint is not provided.
```
avisiedo, line 327: typo: updateServiceBasePath

internal/config/config.go
```diff
@@ -310,6 +311,144 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
 	// Metrics configuration
 	v.Set("metrics.path", clowderConfig.MetricsPath)
 	v.Set("metrics.port", clowderConfig.MetricsPort)
+
+	// If TlsCAPath print an informative message
+	if clowderConfig.TlsCAPath != nil {
+		slog.Info("clowder provides a path for CA, be sure it is in your SSL_CERT_DIR",
+			slog.String("TlsCAPath", *clowderConfig.TlsCAPath),
+		)
```
frasertweedale, line 319: Do I read it correctly, that if the clowder config includes the `TlsCAPath`, then we want to advise the user that `SSL_CERT_DIR` env var should be set, and should include that path?

If that's correct, perhaps we could consider to read and parse `SSL_CERT_DIR` and ensure it - and die (or log a much more serious error level) if we detect a misconfiguration.

build/package/Dockerfile
avisiedo: :+1: thanks mate! updated

deployments/clowder.yaml
avisiedo: :+1: thanks mate! updated!

internal/config/config.go
```diff
@@ -314,6 +315,156 @@ func setClowderConfiguration(v *viper.Viper, clowderConfig *clowder.AppConfig) {
...
+	if clowderConfig.TlsCAPath != nil {
+		envSSLCertDir := os.Getenv("SSL_CERT_DIR")
+		if !checkPathInList(*clowderConfig.TlsCAPath, envSSLCertDir) {
+			slog.Error("clowder provides a path for CA, but the path is not found at SSL_CERT_DIR",
+				slog.String("TlsCAPath", *clowderConfig.TlsCAPath),
+				slog.String("SSL_CERT_DIR", envSSLCertDir),
+			)
+			panic("SSL_CERT_DIR missconfiguration")
+		}
+	}
```
frasertweedale, line 329: OK, thinking aloud even further... if we have TlsCAPath configured, but it is not in `SSL_CERT_DIR` - couldn't we add it to the env var?  (Just seeking comment - I don't want to block this PR on such a change, but maybe it is a way to simplify the configuration in the future).

avisiedo: good question! I think I can get it working! I am not a fun of `init` functions, but I think it will make this work in that way; I mean, as this depends on `clowder` package, the clowder configuration is already loaded, I can check the values, and update the environment variable. Thanks! :+1:

avisiedo:
I hate the way I have checked the change about adding `TlsCAPath` to `SSL_CERT_DIR`, but I didn't find a different method.

- I added several `fmt.Printf` to the function:
```golang
func initSSLCertDir(clowderConfig *clowder.AppConfig) {
	fmt.Printf("initSSLCertDir:1\n")
	fmt.Printf("initSSLCertDir:1:existPath('/cdapp/cdappconfig.json')=%t\n", existsPath("/cdapp/cdappconfig.json"))
	if hasTLSCAPath(clowderConfig) {
		fmt.Printf("initSSLCertDir:2:hasTLSCAPath\n")
		envSSLCertDir := os.Getenv(EnvSSLCertDirectory)
		fmt.Printf("initSSLCertDir:3:envSSLCertDir='%s'\n", envSSLCertDir)
		if !checkPathInList(*clowderConfig.TlsCAPath, envSSLCertDir) {
			fmt.Printf("initSSLCertDir:4:!checkPathInList\n")
			envSSLCertDirItems := make([]string, 0, 10)
			if envSSLCertDir != "" {
				envSSLCertDirItems = append(envSSLCertDirItems, strings.Split(envSSLCertDir, ":")...)
			}
			envSSLCertDirItems = append(envSSLCertDirItems, *clowderConfig.TlsCAPath)
			envSSLCertDir = strings.Join(envSSLCertDirItems, ":")
			fmt.Printf("initSSLCertDir:5:envSSLCertDir='%s'\n", envSSLCertDir)
			if err := os.Setenv(EnvSSLCertDirectory, envSSLCertDir); err != nil {
				panic(err.Error())
			}
			fmt.Printf("initSSLCertDir:6\n")
		}
	}
	fmt.Printf("initSSLCertDir:7\n")
}
```
- I copied a clowder config example from an ephemeral environment.
- I updated to fill `TlsCAPath` field.
- I ran the service as a container locally, mounting the clowder config at `/cdapp` to see the sequence followed as the below:
```golang
initSSLCertDir:1
initSSLCertDir:1:existPath('/cdapp/cdappconfig.json')=true
initSSLCertDir:2:hasTLSCAPath
initSSLCertDir:3:envSSLCertDir='/etc/ssl/certs:/etc/pki/tls/certs:/system/etc/security/cacerts'
initSSLCertDir:4:!checkPathInList
initSSLCertDir:5:envSSLCertDir='/etc/ssl/certs:/etc/pki/tls/certs:/system/etc/security/cacerts:/cdapp/certs'
initSSLCertDir:6
initSSLCertDir:7
```

BTW The unit tests are working.

internal/config/config.go
avisiedo: @frasertweedale , I have updated and checked! and seems to work; I checked in a container launched locally, using a custom clowder configuration (just to check the cert pools); the SSL_CERT_DIR was updated, and I added a self-signed certificate which was loaded (just checked he number of certificates differed from one execution and other); be aware that `SystemCertPool` use in the process (deeper calls) a `once` semantic so the pool of system certificates are loaded only once during the execution. More details on later comments.

frasertweedale: OK, very nice!

internal/config/config.go
```diff
@@ -461,3 +607,45 @@ func readEnv(key string, def string) string {
...
+// in the list of directories.
+// clowderConfig
+func initSSLCertDir(clowderConfig *clowder.AppConfig) {
+	envSSLCertDir := os.Getenv(EnvSSLCertDirectory)
+	if hasTLSCAPath(clowderConfig) {
+		if !checkPathInList(*clowderConfig.TlsCAPath, envSSLCertDir) {
+			envSSLCertDirItems := make([]string, 0, 10)
+			if envSSLCertDir != "" {
+				envSSLCertDirItems = append(envSSLCertDirItems, strings.Split(envSSLCertDir, ":")...)
+			}
```
frasertweedale, line 621: could it be simpler?
```go
if envSSLCertDir != "" {
    envSSLCertDir += ":"
}
envSSLCertDir += *clowderConfig.TlsCAPath
```

This avoids splitting and re-joining the PATH components, and just appends the new element.

internal/config/config.go
```diff
@@ -461,3 +607,45 @@ func readEnv(key string, def string) string {
...
+			envSSLCertDirItems = append(envSSLCertDirItems, *clowderConfig.TlsCAPath)
+			envSSLCertDir = strings.Join(envSSLCertDirItems, ":")
```
frasertweedale, line 623: Could be simpler?

```suggestion
			if envSSLCertDir != "" {
				envSSLCertDir += ":"
			}
			envSSLCertDir += *clowderConfig.TlsCAPath
```

frasertweedale (approved):

PR #324, opened 2024-09-12
feat(HMS-4622): add error response

avisiedo:
This change align the API response for the errors
so it is aligned with the openapi specification
defined for the service. It is made by adding a
default error handler when the echo framework is
initialized.

https://issues.redhat.com/browse/HMS-4622


internal/errors/http/echo/default.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// parseError translate an error into api_public.ErrorInfo and
+// return the first http status code of the error list.
+func parseError(err error) (api_public.ErrorInfo, int) {
+	switch value := err.(type) { //nolint:all
+	case *echo.HTTPError:
+		return api_public.ErrorInfo{
+			Status: strconv.Itoa(value.Code),
+			Title:  value.Error(), /* TODO ?? // value.Message, */
```
pvoborni, line 53: I'd use `Message` as `Code` is put to `Status` and would be redundant in `Title`.  
The implementation of `.Error()` is:
```
func (he *HTTPError) Error() string {
	if he.Internal == nil {
		return fmt.Sprintf("code=%d, message=%v", he.Code, he.Message)
	}
	return fmt.Sprintf("code=%d, message=%v, internal=%v", he.Code, he.Message, he.Internal)
}
```
`Internal`, I believe, is unwrapped as another error.

internal/errors/http/echo/default_test.go
```diff
@@ -0,0 +1,99 @@
...
+						WithDetail("secondary error").
+						Build(),
+					).
+					Build(),
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Log(testCase.Name)
```
pvoborni, line 87: This is a good example of running subtests.  Thus I'd recommend to use the build-in functionality for it: `t.Run(...` 

https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks

pvoborni: This is more an issue of the vscode extension. But isn't it written in the "output" section?

pvoborni: I'd remove the usage of  `t.Log` there. But I'm also thinking if we need to do some cleanup, e.g. stop the server. Or it will be done automatically.

internal/errors/http/echo/default.go
```diff
@@ -0,0 +1,69 @@
...
+		if firstCode == 0 {
+			firstCode = code
+		}
+		if err = errors.Unwrap(err); err == nil {
+			break
+		}
+	}
+	response.Errors = &errList
+	if err = c.JSON(firstCode, &response); err != nil {
+		logger.Error("the error could not be reported in the http response",
```
pvoborni, line 47: Wouldn't message: "failed to send JSON error http response" be more precise here? I.e. "reported" is bit unusual thus confusing here. Also the fact that it failed in making it JSON is missing.

internal/errors/http/echo/default.go
```diff
@@ -0,0 +1,69 @@
...
+			slog.String("err", err.Error()),
+		)
+		return
+	}
+	errList := make([]api_public.ErrorInfo, 0, 10)
+
+	firstCode := 0
+	response := api_public.ErrorResponse{}
+	for {
+		logger.Debug("encapsulating error", slog.String("msg", err.Error()))
```
pvoborni, line 35: I find the word "encapsulating" misleading here. "Encapsulating" is bit a synonym to wrapping. But here we are parsing and later unwrapping (almost the opposite). 

So "parsing error" would be more clear to me here. 

Also isn't he message so low level that it would be worth to be "trace" level?

internal/errors/http/echo/default.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/labstack/echo/v4"
+	"github.com/openlyinc/pointy"
+	api_public "github.com/podengo-project/idmsvc-backend/internal/api/public"
+	app_context "github.com/podengo-project/idmsvc-backend/internal/infrastructure/context"
+)
+
+func DefaultErrorHandler(err error, c echo.Context) {
+	logger := app_context.LogFromCtx(c.Request().Context())
+	if c.Response().Committed {
+		logger.Warn("response already commited",
```
pvoborni, line 25: Isn't it more a debug or trace level message? 

Btw this part and with `HandleError: true,` in log middleware config. The processing is done only once. But with the Warn statement, the same message is reported 3 times in log - in logger middleware, then the "encapsulating error" and then "response already committed".

internal/infrastructure/router/router.go
```diff
@@ -80,9 +80,9 @@ func configCommonMiddlewares(e *echo.Echo, c RouterConfig) {
 		LogStatus: true,
 		LogURI:    true,
 
-		// Forwards error to the global error handler, so it can decide
-		// appropriate status code.
-		HandleError: true,
+		// We need to set this true, to avoid double execution, first by the
+		// logger middleware, second by the echo framework internals.
+		HandleError: false,
```
pvoborni, line 85: The comment says "set this true" but the code is setting it to false. 

But with the `if c.Response().Committed ` in the error handler,  the comment is now incorrect. 

I don't know if `HandleError` should be set to `true` or to `false` as I didn't reproduce a practical difference (all was commited).  But I'd incline to keeping to original code and comment - as there might be a situation where the response was not committed yet and it needs to be set e.g. to internal error.

pvoborni (commented):
I focused on practical testing and how it appears in log. I'd prefer changing debug level of some of the messages to trace or omitting them.  See the comments.

internal/infrastructure/router/router.go
pvoborni: Double execution of error parsing? The error handle is committing the response by calling `JSON()` so the commit check prevents any double execution.

pvoborni: Nevermind. Let's go with what you have.

pvoborni (approved):
LGTM

PR #327, opened 2024-09-16
fix(HMS-1806): closing of http response body in Pendo client

pvoborni:
In one call the body was not closed and in another it was not closed when HTTP status was other than OK.

Now, in both usages, it is "deferred" closed right after checking that response did not error-out (and thus `.Body` should be present).

Note: `// nolint:errcheck` is used as linter complains but the code used is the one mentioned in official docs of `net/http` package.


frasertweedale (approved):

frasertweedale:
Ack.  But I'm curious: what was the impact of these bugs? Memory or FD leak?  Something else?

pvoborni:
When I googled it, the responses were mentioning file descriptors and connections the client reuses (keep-alive).

PR #328, opened 2024-09-17
test: MockServer config for pendo mock

pvoborni:
With this change, `make run` will start the 'service' with Pendo client configured against the MockServer. Thus exercising the pendo_client code.

Interesting links (maybe they should be added to some docs):

- https://www.mock-server.com/mock_server/creating_expectations.html 
- https://app.swaggerhub.com/apis/jamesdbloom/mock-server-openapi/5.14.x#/Expectation


frasertweedale:
Could you please make a mention in the commit message about the removal of the insights inventory mockserver config?  Rationale: it will be easier to find if we ever want to restore that :)

test/mockserver/pendo-expectations.json
```diff
@@ -0,0 +1,50 @@
...
+        "method": "POST",
+        "path": "/data/track",
+        "body": {
+          "type": "JSON",
+          "json": {
+              "type": "track"
+          },
+          "matchType": "ONLY_MATCHING_FIELDS"
+        }
+      },
```
frasertweedale, line 44: Not sure if intended or not, and it probably doesn't matter... I noticed while testing that if we satisfy the integration key and content-type, but fail to satisfy the json body match, mockserver responds 404.  I deduce that lack of match simply "falls off" the end, causing 404, rather than a 400 (which is how I would expect the real server to behave).

Again, it probably doesn't matter for our purposes but I thought I should share this observation - might help somehow in the future :)

frasertweedale:
ACK, modulo requested addition to the commit message.

pvoborni:
Rebased and updated the commit message.

frasertweedale (approved):

PR #329, opened 2024-09-17
tests: host-conf performance tests

pvoborni:
Tests that do load/performance testing of host-conf endpoint.

Prep:
```
$ make compose-clean make-compose-up
```

On Mac adjust soconnmax to more than 1600 to allow more connections on a socket.
```
sudo sysctl -w kern.ipc.somaxconn=2048
```

On Linux this might not be needed, but if it would be, then the key name is different: `net.core.somaxconn`.

Run tests:
```
go clean -testcache && make test-perf
```

done in context of HMS-4654


frasertweedale (approved):
:cowboy_hat_face:

PR #330, opened 2024-09-17
chore: add clarification comment

avisiedo:
This change add some clarification to keep the potential values closer to the method definition, so it is more accessible for the developer when implementing the smoke and integration tests.

### DIFF
--- a/internal/test/smoke/suite_base_test.go
+++ b/internal/test/smoke/suite_base_test.go
@@ -168,6 +168,14 @@ func (s *SuiteBase) TearDownTest() {
 	s.wg.Wait()
 }
 
+// As is a helper function that makes it easy to select
+// the XRHID and RBAC profiles to be used with the
+// following requests.
+// profiles could be:
+//
+//   - for rbac: RBACSuperAdmin, RBACAdmin, RBACReadOnly, RBACNoPermis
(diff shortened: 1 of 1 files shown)


internal/test/smoke/suite_base_test.go
```diff
@@ -168,6 +168,13 @@ func (s *SuiteBase) TearDownTest() {
 	s.wg.Wait()
 }
 
+// As is a helper function that make easy to select
+// the XRHID and RBAC profiles to be used with the
+// following requests.
+// profiles could be:
+//
+// - for rbac: RBACSuperAdmin, RBACAdmin, RBACReadOnly, RBACNoPermis
+// - for XRHID: XRHIDNothing, XRHIDNone, XRHIDUser, XRHIDSystem
```
pvoborni, line 177: What about `XRHIDServiceAccount` ?

internal/test/smoke/suite_base_test.go
```diff
@@ -168,6 +168,13 @@ func (s *SuiteBase) TearDownTest() {
 	s.wg.Wait()
 }
 
+// As is a helper function that make easy to select
```
pvoborni, line 171: Nit: isn't the correct grammar: "makes it easy " instead of "make easy"?

pvoborni (approved):
LGTM

PR #332, opened 2024-09-20
refactor(HMS-4648): Inject config.Config

avisiedo:
This refactor inject config.Config during the initialization, so
every component is using the same instance, so the code only
retrieve the value from the singleton at a unique place.

This is necessary for the Limit and Timeout tests, as the initial
configuration is override for every test executed for the different
use cases.

During the refactor, `RouterConfig` was not required anymore, which
is clean-up from the repository.

Depended by: https://github.com/podengo-project/idmsvc-backend/pull/334

https://issues.redhat.com/browse/HMS-4648


internal/infrastructure/router/router_test.go
```diff
@@ -151,121 +68,92 @@ func TestLoggerSkipperWithPaths(t *testing.T) {
...
-	assert.Panics(t, func() {
-		_ = NewRouterWithConfig(e, badConfig)
+	reg := prometheus.NewRegistry()
+	metrics := metrics.NewMetrics(reg)
+	assert.NotPanics(t, func() {
+		NewRouterWithConfig(e, cfg, app, metrics)
 	})
+}
+
+func TestNewRouterWithConfig(t *testing.T) {
```
pvoborni, line 102: Nit: This test seems to cover the same thing as the last part of `TestCheckNewRouterWithConfig`

internal/infrastructure/router/router_test.go
```diff
@@ -151,121 +68,92 @@ func TestLoggerSkipperWithPaths(t *testing.T) {
...
-	if err != nil {
-		panic(err)
-	}
-	rbac := client_rbac.New(cfg.Clients.RbacBaseURL, rbacClient)
-	pendo := client_pendo.NewClient(cfg)
-	// Create application handlers
-	app := handler_impl.NewHandler(cfg, db, metrics, inventory, rbac, pendo)
+	oldMetricsPath := cfg.Metrics.Path
+	cfg.Metrics.Path = ""
+	assert.PanicsWithValue(t, "'app' is nil", func() {
```
pvoborni, line 127: `'app' is nil` should probably be `'metrics' is nil`

internal/infrastructure/router/private_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func TestNewGroupPrivate(t *testing.T) {
+	e := echo.New()
+	require.NotNil(t, e)
+	handlers := api_private.NewServerInterface(t)
+	require.NotNil(t, handlers)
+	assert.NotPanics(t, func() {
+		require.NotNil(t, newGroupPrivate(e.Group("/private"), handlers))
+	})
+	require.NotNil(t, e)
```
pvoborni, line 20: NIt: redundant `require.NotNil(t, e)` - is present earlier.

internal/infrastructure/router/public_test.go
```diff
@@ -437,3 +387,26 @@ func TestSkipperMixedPredicate(t *testing.T) {
 		assert.Equal(t, testCase.Expected, result)
 	}
 }
+
+func TestGuardNewGroupPublic(t *testing.T) {
```
pvoborni, line 391: usually guard tests use `assert.PanicsWithValue` but here is only 4 times:   `assert.Panics` so it is not clear if the right/expected panic is caught.

pvoborni (commented):
This change simplifies some stuff. I like it.

Added couple comments about smaller issues in tests.

PR #333, opened 2024-09-23
feat(HMS-4757): measure header and body request sizes

avisiedo:
This change add metrics to retrieve the size for header and body requests, by adding the necessary changes to metrics structure, and the behavior to store the metrics at the metrics middleware.

https://issues.redhat.com/browse/HMS-4757

### DIFF
--- a/internal/infrastructure/middleware/metrics.go
+++ b/internal/infrastructure/middleware/metrics.go
@@ -2,11 +2,14 @@ package middleware
 
 import (
 	"errors"
+	"log/slog"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/labstack/echo/v4"
 	echo_middleware "github.com/labstack/echo/v4/middleware"
+	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/context"
 	"github.com/podengo-project/idmsvc-backend/internal/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -40,6 +43,8 @@ func MetricsMiddlewareWithConfig(config *MetricsConfig) echo.MiddlewareFunc {
 
 			err := next(ctx)
 
+			logger := context.LogFromCtx(ctx.Request().Context())
+
 			method := ctx.Request().Method
 			path := MatchedRoute(ctx)
 			status := ctx.Response().Status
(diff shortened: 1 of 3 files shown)


internal/infrastructure/middleware/metrics.go
```diff
@@ -50,6 +55,29 @@ func MetricsMiddlewareWithConfig(config *MetricsConfig) echo.MiddlewareFunc {
...
+			bodySize := float64(ctx.Request().ContentLength)
+			config.Metrics.HTTPRequestBodySize.WithLabelValues(statusStr, method, path).Observe(bodySize)
+			logger.Debug("measured http_request_body_size",
+				slog.String("status", statusStr),
+				slog.String("method", method),
+				slog.String("path", path),
+				slog.Float64("size", bodySize),
+			)
 
 			config.Metrics.HTTPRequestDuration.WithLabelValues(statusStr, method, path).Observe(time.Since(start).Seconds())
```
pvoborni, line 83: I was thinking about the logging here. We log the body and header size but don't log the duration. What would you think about single log statement covering all?

I.e. the log output in debug mode would be more concise but having more info.

```diff
diff --git a/internal/infrastructure/middleware/metrics.go b/internal/infrastructure/middleware/metrics.go
index 202870a..b904496 100644
--- a/internal/infrastructure/middleware/metrics.go
+++ b/internal/infrastructure/middleware/metrics.go
@@ -56,15 +56,10 @@ func MetricsMiddlewareWithConfig(config *MetricsConfig) echo.MiddlewareFunc {
                        }
                        statusStr := strconv.Itoa(status)
                        headerBuf := strings.Builder{}
+                       headerSize := 0.0
                        if errHeaderBuf := ctx.Request().Header.Write(&headerBuf); errHeaderBuf == nil {
-                               headerSize := float64(len(headerBuf.String()))
+                               headerSize = float64(len(headerBuf.String()))
                                config.Metrics.HTTPRequestHeaderSize.WithLabelValues(statusStr, method, path).Observe(headerSize)
-                               logger.Debug("measured http_request_header_size",
-                                       slog.String("status", statusStr),
-                                       slog.String("method", method),
-                                       slog.String("path", path),
-                                       slog.Float64("size", headerSize),
-                               )
                        } else {
                                logger.Warn("writing headers in string buffer",
                                        slog.String("err", errHeaderBuf.Error()),
@@ -72,15 +67,19 @@ func MetricsMiddlewareWithConfig(config *MetricsConfig) echo.MiddlewareFunc {
                        }
                        bodySize := float64(ctx.Request().ContentLength)
                        config.Metrics.HTTPRequestBodySize.WithLabelValues(statusStr, method, path).Observe(bodySize)
-                       logger.Debug("measured http_request_body_size",
+
+                       durationSeconds := time.Since(start).Seconds()
+                       config.Metrics.HTTPRequestDuration.WithLabelValues(statusStr, method, path).Observe(durationSeconds)
+
+                       logger.Debug("measured http request",
                                slog.String("status", statusStr),
                                slog.String("method", method),
                                slog.String("path", path),
-                               slog.Float64("size", bodySize),
+                               slog.Float64("body_size", bodySize),
+                               slog.Float64("header_size", headerSize),
+                               slog.Float64("duration_seconds", durationSeconds),
                        )
 
-                       config.Metrics.HTTPRequestDuration.WithLabelValues(statusStr, method, path).Observe(time.Since(start).Seconds())
-
                        return err
                }
        }
```

pvoborni (commented):
I was thinking if the calculation of header size is unnecessary complexity/computation. But it's probably fine. 

Added a suggestion about logging.

internal/infrastructure/middleware/metrics.go
pvoborni: ok, I'd still include even the time as I think the one log statement doesn't play a big role (there might be some stuff after it anyway in the echo framework). 

But from the PoV of this PR it LGTM. So will approve.

pvoborni (approved):
LGTM

PR #334, opened 2024-09-23
refactor: remove inventory client

avisiedo:
This change clean-up the inventory component from
the repository, which is not used and not necessary.

This change is not required, but it was observed
during the refactor to use the injected configuration
and remove `RouterConfig` as it was a pending
task, and as it was being modified the intialization
functions it was considered a good moment for the
clean-up. However, this has dependency on the previous
PR because conflicts on some lines that both modifies.

Depends on: https://github.com/podengo-project/idmsvc-backend/pull/332

Dpended by: https://github.com/podengo-project/idmsvc-backend/pull/331

https://issues.redhat.com/browse/HMS-4648


internal/handler/impl/application_test.go
```diff
@@ -6,38 +6,59 @@ import (
...
+
+	m := &metrics.Metrics{}
+	assert.PanicsWithValue(t, "'rbac' is nil", func() {
+		NewHandler(cfg, gormDB, m, nil, nil)
 	})
-	cfg := test.GetTestConfig()
+
+	rbacClient := rbac.NewRbac(t)
+	assert.PanicsWithValue(t, "'pendo' is nil", func() {
+		NewHandler(cfg, gormDB, &metrics.Metrics{}, rbacClient, nil)
```
pvoborni, line 42: nit: shouldn't this use `m` instead of `&metrics.Metrics{}` - to have consistent style. Same comment applies also for next call.

pvoborni (commented):
The removal of inventory LGTM.  I'd OK even with not fixing the one comment.

pvoborni (approved):
LGTM. 

Btw I did a `git grep inventory` and there are still some remaining stuff, like a config value for inventory endpoint.  But we can remove them in separate PR and not block the other PRs which depends on this.

pvoborni:
Still LGTM.

PR #339, opened 2024-10-02
fix: avoid panic due to unset XRHID on invalid path

frasertweedale:
If the HTML resource path is unrecognised, the EnforceIdentity middlewares all get skipped.  This causes the XRHID field in the context to be uninitialised.  As a consequence, the RBAC middleware panics because it expects the field to be set.

To avoid this issue, we create a new middleware which always sets the XHRID into the request context.  This middleware runs after the FakeIdentity middleware (if used), and before the EnforceIdentity and RBAC middlewares.

The EnforceIdentity middleware is simplified as a result; it no longer needs to parse the XRHID and can read it from the context.

### DIFF
--- a/internal/infrastructure/middleware/enforce_identity_test.go
+++ b/internal/infrastructure/middleware/enforce_identity_test.go
@@ -55,6 +55,7 @@ func helperNewEchoEnforceIdentity(m echo.MiddlewareFunc) *echo.Echo {
 	}
 	e.Use(ContextLogConfig(&LogConfig{}))
 	e.Use(CreateContext())
+	e.Use(ParseXRHIDMiddlewareWithConfig(&ParseXRHIDMiddlewareConfig{}))
 	e.Use(m)
 	e.Add("GET", testPath, h)
 
@@ -105,8 +106,8 @@ func TestEnforceIdentity(t *testing.T) {
 			Name:  header.HeaderXRHID + " header not present",
 			Given: nil,
 			Expected: TestCaseExpected{
-				Code: http.StatusBadRequest,
-				Body: "{\"message\":\"Bad Request\"}\n",
+				Code: http.StatusUnauthorized,
+				Body: "{\"message\":\"Unauthorized\"}\n",
 			},
 		},
 		{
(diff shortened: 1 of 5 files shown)


internal/infrastructure/middleware/enforce_identity_test.go
```diff
@@ -240,33 +241,52 @@ func TestEnforceIdentitySkipper(t *testing.T) {
...
+						Name:      "test-fail-predicate",
+						Predicate: helperCreatePredicate("test-fail-predicate"),
+					},
+				},
 			},
 		),
 	)
 	res = httptest.NewRecorder()
 	req = httptest.NewRequest(http.MethodGet, "/test", nil)
+	req.Header.Add(header.HeaderXRHID, header.EncodeXRHID(&userXRHID))
```
pvoborni, line 265: This line does not match the `e` comment: `// When skipper return false, as no x-rh-identity provided, will return unauthorized` I.e. the identity is provided. 

Btw the test seems to still pass when the modifications on lines 254-259 and 265 are removed.

pvoborni (commented):
Added a small comment. 

Also a Q if we should add a test to test the actually use case - the wrong path and verify that we got 404 instead of 500.

internal/infrastructure/middleware/enforce_identity_test.go
frasertweedale: > This line does not match the `e` comment: `// When skipper return false, as no x-rh-identity provided, will return unauthorized` I.e. the identity is provided.

Good catch.  I updated the comments.

> Btw the test seems to still pass when the modifications on lines 254-259 and 265 are removed.

Yes, but it will pass for the wrong reason :)  Due to the changes, this would pass because the `ParseXRHID` middleware returns 401 because it does not see the `X-RH-Identity` header.  We want it to pass because the `EnforceIdentity` predicate fails (because the skipper *doesn't* skip the middleware).

frasertweedale:
> Also a Q if we should add a test to test the actually use case - the wrong path and verify that we got 404 instead of 500.

I can't think of a good (and lightweight) way to do it in the unit tests.  Maybe the integration tests is a better place.

> In any case, I've checked ^^ manually and it works.

Thanks!

frasertweedale:
Added a smoke test :)

pvoborni (approved):
LGTM

PR #340, opened 2024-10-02
fix(HMS-3393): reject domain update if server RHSM ID not in server list

pvoborni:
So that a server cannot DOS itself.

The patch is also fixing test flakiness when HCCUpdate == false by setting the HCCUpdate to true for register domain flows.

The check is the same as for ensureSubscriptionManagerIDAuthorizedToUpdate but the error is different (BadRequest vs Unauthorized).


frasertweedale (approved):
LGTM

PR #341, opened 2024-10-03
fix(HMS-4789): panic on /api/idmsvc/v1/* not found route

avisiedo:
This change do an early return in the global
middleware chain, to avoid panics for the not
found routes for the API endpoints.

https://issues.redhat.com/browse/HMS-4789


avisiedo:
closing as override by https://github.com/podengo-project/idmsvc-backend/pull/347

PR #342, opened 2024-10-03
fix(HMS-4769): log configuration on startup

pvoborni:
Log also secrets but obfuscate them so that it is visible if they are set or not but not their value.

Setting as draft to get initial feedback. Qs:

- is it worth the time to implement tests for this. IMHO the only useful test would be to check that secrets are obfuscated. But to test it, it would need something like: https://pkg.go.dev/github.com/thejerf/slogassert#Handler.Assert 
- is the place where it is logged good? I.e before or after logger initialization? 

Example output:

> time=2024-10-03T18:02:34.477+02:00 level=INFO msg=idmscv-backend Version=(devel) Commit=5e87078ba927d1d626e84a7f873527abb357b4bf CommitTime=2024-10-03T06:00:45Z Dirty=true
> time=2024-10-03T18:02:34.478+02:00 level=INFO msg=idmsvc-backend Logging.Console=true Logging.Type=null Logging.Level=debug Logging.Location=false
> time=2024-10-03T18:02:34.478+02:00 level=NOTICE msg="Logging successfully initialized"
> time=2024-10-03T18:02:34.478+02:00 level=INFO msg=Configuration Web.Port=8000 Database.Host=localhost Database.Port=5432 Database.User=idmsvc-user Database.Password=*** Database.Name=idmsvc-db Database.CACertPath="" Database.MaxOpenConns=30 Logging.Level=debug Logging.Console=true Logging.Location=false Logging.Type=null Logging.Cloudwatch.Region="" Logging.Cloudwatch.Key="" Logging.Cloudwatch.Secret="" Logging.Cloudwatch.Session="" Logging.Cloudwatch.Group="" Logging.Cloudwatch.Stream=pvoborni-mac Metrics.Path=/metrics Metrics.Port=9000 Clients.RbacBaseURL=http://localhost:8020/api/rbac/v1 Clients.PendoBaseURL=http://localhost:8010 Clients.PendoAPIKey=*** Clients.PendoTrackEventKey=*** Clients.PendoRequestTimeoutSecs=0 Application.Name=idmsvc Application.PathPrefix=/api/idmsvc/v1 Application.TokenExpirationTimeSeconds=7200 Application.HostconfJwkValidity=2160h0m0s Application.HostconfJwkRenewalThreshold=720h0m0s Application.PaginationDefaultLimit=10 Application.PaginationMaxLimit=100 Application.AcceptXRHFakeIdentity=true Application.ValidateAPI=true Application.MainSecret=*** Application.EnableRBAC=true Application.IdleTimeout=5m0s Application.ReadTimeout=3s Application.WriteTimeout=3s Application.SizeLimitRequestHeader=32768 Application.SizeLimitRequestBody=131072
> time=2024-10-03T18:02:34.524+02:00 level=DEBUG msg="Printing metrics routes"

### DIFF
--- a/cmd/service/main.go
+++ b/cmd/service/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"log/slog"
 	"os"
 	"os/signal"
 	"sync"
@@ -47,6 +48,7 @@ func main() {
 	logger.LogBuildInfo("idmscv-backend")
 	cfg := config.Get()
 	logger.InitLogger(cfg)
+	cfg.Log(slog.Default())
 	db := datastore.NewDB(cfg)
 	defer datastore.Close(db)
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -509,6 +509,75 @@ func Load(cfg *Config) *viper.Viper {
 	return v
 }
 
+// Log logs the configuration
+func (c *Config) Log(logger *slog.Logger) {
+	logger.Info(
+		"Configuration",
+		slog.Group("Web",
+			slog.Int("Port", int(c.Web.Port)),
+		),
+		slog.Group("Database",
+			slog.String("Host", c.Database.Host),
+			slog.Int("Port", c.Database.Port),
+			slog.String("User", c.Database.User),
+			slog.String("Password", obfuscateSecret(c.Database.Password)),
+			slog.String("Name", c.Database.Name),
+			slog.String("CACertPath", c.Database.CACertPath),
+			slog.Int("MaxOpenConns", c.Database.MaxOpenConns),
+		),
+		slog.Group("Logging",
+			slog.String("Level", c.Logging.Level),
+			slog.Bool("Console", c.Logging.Console),
+			slog.Bool("Location", c.Logging.Location),
+			slog.String("Type", c.Logging.Type),
+			slog.Group("Cloudwatch",
+				slog.String("Region", c.Logging.Cloudwatch.Region),
+				slog.String("Key", c.Logging.Cloudwatch.Key),
+				slog.String("Secret", obfuscateSecret(c.Logging.Cloudwatch.Secret)),
+				slog.String("Session", c.Logging.Cloudwatch.Session),
+				slog.String("Group", c.Logging.Cloudwatch.Group),
+				slog.String("Stream", c.Logging.Cloudwatch.Stream),
+			),
+		),
+		slog.Group("Metrics",
+			slog.String("Path", c.Metrics.Path),
+			slog.Int("Port", c.Metrics.Port),
+		),
+		slog.Group("Clients",
+			slog.String("RbacBaseURL", c.Clients.RbacBaseURL),
+			slog.String("PendoBaseURL", c.Clients.PendoBaseURL),
+			slog.String("PendoAPIKey", obfuscateSecret(c.Clients.PendoAPIKey)),
+			slog.String("PendoTrackEventKey", obfuscateSecret(c.Clients.PendoTrackEventKey)),
+			slog.Int("PendoRequestTimeoutSecs", c.Clients.PendoRequestTimeoutSecs),
+		),
+		slog.Group("Application",
+			slog.String("Name", c.Application.Name),
+			slog.String("PathPrefix", c.Application.PathPrefix),
+			slog.Int("TokenExpirationTimeSeconds", c.Application.TokenExpirationTimeSeconds),
+			slog.Duration("HostconfJwkValidity", c.Application.HostconfJwkValidity),
+			slog.Duration("HostconfJwkRenewalThreshold", c.Application.HostconfJwkRenewalThreshold),
+			slog.Int("PaginationDefaultLimit", c.Application.PaginationDefaultLimit),
+			slog.Int("PaginationMaxLimit", c.Application.PaginationMaxLimit),
+			slog.Bool("AcceptXRHFakeIdentity", c.Application.AcceptXRHFakeIdentity),
+			slog.Bool("ValidateAPI", c.Application.ValidateAPI),
+			slog.String("MainSecret", obfuscateSecret(c.Application.MainSecret)),
+			slog.Bool("EnableRBAC", c.Application.EnableRBAC),
+			slog.Duration("IdleTimeout", c.Application.IdleTimeout),
+			slog.Duration("ReadTimeout", c.Application.ReadTimeout),
+			slog.Duration("WriteTimeout", c.Application.WriteTimeout),
+			slog.Int("SizeLimitRequestHeader", c.Application.SizeLimitRequestHeader),
+			slog.Int("SizeLimitRequestBody", c.Application.SizeLimitRequestBody),
+		),
+	)
+}
+
+func obfuscateSecret(value string) string {
+	if value == "" {
+		return ""
+	}
+	return "***"
+}
+
 func reportError(err error) {
 	for _, err := range err.(validator.ValidationErrors) {
 		slog.Error(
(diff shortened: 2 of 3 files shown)


cmd/service/main.go
```diff
@@ -47,6 +48,7 @@ func main() {
 	logger.LogBuildInfo("idmscv-backend")
 	cfg := config.Get()
 	logger.InitLogger(cfg)
+	cfg.Log(*slog.Default())
```
avisiedo, line 51: Q: About the place to call to this method:

- Do we want to have this information at clowdwatch? or only in the pod? for the second I have one uncertainty: if the pod logs could be recovered if the pod panic during initialization time, before call the config dump.
- Arguments:
  - if the pod logs could be recovered in that case **and** we don't want this in cloudwatch, I would say to move the call just after `cfg := config.Get()`.
  - In other case, the place where it is called I would agree with it.

internal/config/config.go
```diff
@@ -509,6 +509,75 @@ func Load(cfg *Config) *viper.Viper {
 	return v
 }
 
+// Log logs the configuration
+func (c *Config) Log(logger slog.Logger) {
```
avisiedo, line 513: I suggest to divide the method in smaller private ones, so this method could looks like:

```go
func (c *Config) Log(logger slog.Logger) {
	if c == nil {
		panic("'c' is a nil &Config{} pointer")
	}
	logger.Info(
		"Configuration",
		slog.Bool("Loaded", c.Loaded),
		c.logGroupWeb(&c.Web),
		c.logGroupDatabase(&c.Database),
		c.logGroupLogging(&c.Logging),
		c.logGroupMetrics(&c.Metrics),
		c.logGroupClients(&c.Clients),
		c.logGroupApplication(&c.Application),
		// No logGroupSecrets() is intentional
	)
}
```

And for instance for the group web could looks like:

```go
func (c *Config) logGroupWeb(data *Web) slog.Attr {
	return slog.Group("Web",
		slog.Int("Port", int(data.Port)),
	)
}
```

In this way, the unit tests would be easier to accommodate and independent of the format; we could check for every private method, that the `slog.Attr` structure returned fit our expectations which is independent of the logging format; for instance for logGroupWeb it could looks like:

```go
func TestGroupWeb(t *testing.T) {
	data := &Config{
		Web: Web{
			Port: 8000,
		},
	}
	assert.Equal(t,
		slog.Group("Web",
			slog.Int("Port", 8000),
		),
		data.logGroupWeb(&data.Web),
	)
}
```

In that way we could cover more lines with the unit tests, without consider the output format complexity of the log.

For the log format, I had a vague memory in the past; I have found this unit test, but I agree that check the output format as unnecessary:

https://github.com/podengo-project/idmsvc-backend/blob/main/internal/infrastructure/logger/meta_handler_test.go#L122

> Dividing the method in smaller one, we could get coverage for almost everything without enter on the output format details.

internal/config/config.go
```diff
@@ -509,6 +509,75 @@ func Load(cfg *Config) *viper.Viper {
...
+			slog.Int("MaxOpenConns", c.Database.MaxOpenConns),
+		),
+		slog.Group("Logging",
+			slog.String("Level", c.Logging.Level),
+			slog.Bool("Console", c.Logging.Console),
+			slog.Bool("Location", c.Logging.Location),
+			slog.String("Type", c.Logging.Type),
+			slog.Group("Cloudwatch",
+				slog.String("Region", c.Logging.Cloudwatch.Region),
+				slog.String("Key", c.Logging.Cloudwatch.Key),
```
avisiedo, line 535: uncertainty: should this field be obfuscated too?

internal/config/config.go
```diff
@@ -509,6 +509,75 @@ func Load(cfg *Config) *viper.Viper {
...
+		),
+		slog.Group("Metrics",
+			slog.String("Path", c.Metrics.Path),
+			slog.Int("Port", c.Metrics.Port),
+		),
+		slog.Group("Clients",
+			slog.String("RbacBaseURL", c.Clients.RbacBaseURL),
+			slog.String("PendoBaseURL", c.Clients.PendoBaseURL),
+			slog.String("PendoAPIKey", obfuscateSecret(c.Clients.PendoAPIKey)),
+			slog.String("PendoTrackEventKey", obfuscateSecret(c.Clients.PendoTrackEventKey)),
```
avisiedo, line 550: uncertainty: should this field be obfuscated too?

avisiedo (commented):
Thanks for the change, this will make life easier in the future.

I have some suggestions that could allow unit tests without consider the format of the log; and some questions related with uncertainty.

BTW I verified the change locally and using the json format print out something like the below (output after use `jq` for prettier format):

```json
{
  "time": "2024-10-04T10:06:34.073118125+02:00",
  "level": "INFO",
  "source": {
    "function": "github.com/podengo-project/idmsvc-backend/internal/config.(*Config).Log",
    "file": "/home/avisiedo/work/hms/hmsidm/idmsvc-backend/internal/config/config.go",
    "line": 514
  },
  "msg": "Configuration",
  "Web": {
    "Port": 8000
  },
  "Database": {
    "Host": "localhost",
    "Port": 5432,
    "User": "idmsvc-user",
    "Password": "***",
    "Name": "idmsvc-db",
    "CACertPath": "",
    "MaxOpenConns": 30
  },
  "Logging": {
    "Level": "info",
    "Console": false,
    "Location": true,
    "Type": "null",
    "Cloudwatch": {
      "Region": "",
      "Key": "",
      "Secret": "",
      "Session": "",
      "Group": "",
      "Stream": "avisiedo-thinkpadp1gen2.rmtes.csb"
    }
  },
  "Metrics": {
    "Path": "/metrics",
    "Port": 9000
  },
  "Clients": {
    "RbacBaseURL": "http://localhost:8020/api/rbac/v1",
    "PendoBaseURL": "http://localhost:8010",
    "PendoAPIKey": "***",
    "PendoTrackEventKey": "***",
    "PendoRequestTimeoutSecs": 10
  },
  "Application": {
    "Name": "idmsvc",
    "PathPrefix": "/api/idmsvc/v1",
    "TokenExpirationTimeSeconds": 7200,
    "HostconfJwkValidity": 7776000000000000,
    "HostconfJwkRenewalThreshold": 2592000000000000,
    "PaginationDefaultLimit": 10,
    "PaginationMaxLimit": 100,
    "AcceptXRHFakeIdentity": true,
    "ValidateAPI": true,
    "MainSecret": "***",
    "EnableRBAC": true,
    "IdleTimeout": 300000000000,
    "ReadTimeout": 3000000000,
    "WriteTimeout": 3000000000,
    "SizeLimitRequestHeader": 32768,
    "SizeLimitRequestBody": 65535
  }
}
```

internal/config/config.go
pvoborni: I don't think so. It is used for `AccessKeyID` which is sort of "username". 

Similar Q: https://security.stackexchange.com/questions/187992/is-an-aws-access-key-id-a-secret

internal/config/config.go
pvoborni: Meaning `PendoTrackEventKey` ? It is already, isn't it?

internal/config/config.go
pvoborni: I don't see a lot of benefits in such test as it is practically copying the implementation and thus would break with small implementation change - i.e. it is testing implementation and not the behavior.

Did a small research and I've added a test that verifies at least one attr from eah section + that all secrets values are not present.

internal/config/config.go
avisiedo: Yes, I read it wrongly! thanks :+1:

cmd/service/main.go
avisiedo: I don't think this is something for blocking the pr and we can think about it later; closing the conversation; if we change our mind we can read the pr and rethink about it.

avisiedo (approved):
LGTM thanks!

PR #347, opened 2024-10-08
feat (HMS-4789): Version of #341 that does not revert #339

frasertweedale:
This PR rebases #341, but excludes the commits that revert #339.

 https://issues.redhat.com/browse/HMS-4789

### DIFF
--- a/internal/infrastructure/middleware/not_found.go
+++ b/internal/infrastructure/middleware/not_found.go
@@ -1,0 +1,28 @@
+package middleware
+
+import (
+	"log/slog"
+	"strings"
+
+	"github.com/labstack/echo/v4"
+	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/context"
+)
+
+// NotFound middleware is a workaround to avoid the middleware chain
+// is executed when the path is not found, because at the moment of
+// writting this, the not found paths are matching a default handler.
+//
+// This workaround evoke an early return by calling the NotFoundHandler
+func NotFound() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			path := c.Path()
+			if path == "" || strings.HasSuffix(path, "/*") {
+				logger := context.LogFromCtx(c.Request().Context())
+				logger.Error("not found", slog.String("path", path))
+				return echo.NotFoundHandler(c)
+			}
+			return next(c)
+		}
+	}
+}
--- a/internal/test/smoke/miscelanea_test.go
+++ b/internal/test/smoke/miscelanea_test.go
@@ -1,0 +1,135 @@
+package smoke
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/openlyinc/pointy"
+	"github.com/podengo-project/idmsvc-backend/internal/api/public"
+	"github.com/podengo-project/idmsvc-backend/internal/infrastructure/datastore"
+	builder_api "github.com/podengo-project/idmsvc-backend/internal/test/builder/api"
+	mock_pendo "github.com/podengo-project/idmsvc-backend/internal/test/mock/interface/client/pendo"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type SuiteMiscelanea struct {
+	SuiteBase
+	token  *public.DomainRegTokenResponse
+	domain *public.Domain
+}
+
+func (s *SuiteMiscelanea) SetupTest() {
+	s.PendoClient = mock_pendo.NewPendo(s.T())
+	s.SuiteBase.SetupTest()
+}
+
+func (s *SuiteMiscelanea) prepareDomainIpaCreate(t *testing.T) {
+	s.As(XRHIDUser)
+	token, err := s.CreateToken()
+	require.NoError(t, err)
+	require.NotNil(t, token)
+	require.NotEqual(t, "", token.DomainToken)
+	s.token = token
+}
+
+func (s *SuiteMiscelanea) prepareDomainIpa(t *testing.T) {
+	var err error
+
+	// Add key to the database
+	t.Log("Adding key")
+	hcdb := datastore.NewHostconfJwkDb(s.Config)
+	hcdb.ListKeys()
+	hcdb.Purge()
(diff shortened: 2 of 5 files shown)


avisiedo:
My bad, I thought that we would want to revert the previous change.

lgtm to merge without revert, but it needs a review for the new commits.

internal/test/smoke/miscelanea_test.go
```diff
@@ -0,0 +1,135 @@
...
+			WithAutoEnrollmentEnabled(pointy.Bool(true)).
+			Build())
+	require.NoError(t, err)
+	require.NotNil(t, s.domain)
+}
+
+// TestBugHmsXXXXNotFound check a condition when a path
+// that is not found for /api/idmsvc/domains/v1/host-conf/:inventory_id
+// where the API should return with 404 status code.
+func (s *SuiteMiscelanea) TestBugHmsXXXXNotFound() {
```
frasertweedale, line 94: Shall we replace the `XXXX` with `4789`?

internal/infrastructure/middleware/not_found.go
```diff
@@ -0,0 +1,28 @@
...
+// NotFound middleware is a workaround to avoid the middleware chain
+// is executed when the path is not found, because at the moment of
+// writting this, the not found paths are matching a default handler.
+//
+// This workaround evoke an early return by calling the NotFoundHandler
+func NotFound() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			path := c.Path()
+			if path == "" || strings.HasSuffix(path, "/*") {
```
frasertweedale, line 20: I couldn't find any documentation (echo?) that explains how `c.Path()` is set, based on the route processing.  In particular, under what circumstances would its value be:
- empty, or
- suffixed with `"/*"`

Can you point to the docs or the echo source code that explain this behaviour?  The code here "looks right" but I am missing a concrete understanding of *why*.

This is **not a blocker** to merging the PR, because the tests are passing :)  But I want to complete my understanding.

frasertweedale (commented):
The code LGTM but I have a gap to complete my understanding of it, and one suggestion to replace a missing ticket number.

internal/test/smoke/miscelanea_test.go
frasertweedale: Lol, just remembered that this is my PR so I can update it myself :)

internal/infrastructure/middleware/not_found.go
avisiedo: I based my changes on observations and experiments rather than official documentation.

But looking at the documentation, there are something that worry me about the change, and it is that the route allow to use wildcards (see: https://echo.labstack.com/docs/routing#match-any--wildcard). So, this change is working because we are not using wildcards on our route definition. We have a singularity, and this change take advantage of it. In other words, as we are not using `group.Any(...)` to define some route, this change is working.

Looking the code of the framework closely, I can see that when a group has middlewares associated, it is called RouteNotFound for paths = `""` and `"/*"` (to hanlde the 404 response).

Links of interest:
* https://echo.labstack.com/docs/routing#match-any--wildcard
* https://github.com/labstack/echo/blob/master/group.go#L22

avisiedo: I suggest to update title by: `feat(HMS-4789): Version of #341 that does not revert #339 #347`, and add link to the ticket in the PR body: https://issues.redhat.com/browse/HMS-4789

avisiedo:
I suggest to update title by: `feat(HMS-4789): Version of #341 that does not revert #339 #347`, and add link to the ticket in the PR body: https://issues.redhat.com/browse/HMS-4789

internal/infrastructure/middleware/not_found.go
frasertweedale: Yeah, I don't fully comprehend all the details of echo's routing system.  What is clear from the comments in the `group.go` code you linked is that it is echo's intended behaviour to unconditionally execute group middlewares **before** invoking a NotFoundHandler.  This is _related to_ the cause of the 500 bug.  Our middlewares did not contemplate that they would be executed with invalid paths.  #341 was one way of resolving that.

This PR is (complementary and compatible) way of resolving that - by introducing a _middleware_ to detect the invalid path and short-circuit with a 404.  I am not 100% confident that the change does what we want, because of uncertainty about exactly how echo sets the `Path()`.  But I'm pretty confident that it won't make things worse.

Do you think we should proceed with the change, given that #341 has been merged and verified to fix the issue?
Let me know what you think (ping @pvoborni).

If we decide not to proceed, we should extract and land the XHRID cleanup separately.

avisiedo: The `ctx.Path()` value is set by the method `Router.Find(...)` ; below some links of interest:
* `ServeHTTP` method implementation: https://github.com/labstack/echo/blob/master/echo.go#L653
* `Find` method implementation: https://github.com/labstack/echo/blob/master/router.go#L542

After the process of matching the request URL with the router information, at the end is set the path in the request context.

How the matching process is complex to find out from the source code (the method is not simple); but we can get a high level view from the echo documentation at: https://echo.labstack.com/docs/routing#path-matching-order

> I have not found yet anything that say the `Path == ""` or `Path == "/*"` when no route is found. So still the resulting change is based on the observations.

pvoborni: From my PoV, fix done in #341 was sufficient. I.e. I didn't see a strong need to improve it further but I was OK with it. But if the recent finding shows that this implementation is diverging from intended behavior of echo framework then I'd rather not do it - to not make us less predictable/consistent.

+1 to doing the XHRID cleanup.

avisiedo: I have extracted the XRHID cleanup on this PR https://github.com/podengo-project/idmsvc-backend/pull/349 / so the commit can be removed from this PR

I have closed https://github.com/podengo-project/idmsvc-backend/pull/341 as the result of the changes will be on this pr.

frasertweedale: Based on the feedback, closing this PR too.  Thanks @avisiedo for the proposal and discussion - I know more about echo framework internals as a result :)

PR #356, opened 2024-10-16
fix(HMS-3803): add WithDomainName to Domain builder

avisiedo:
Small fix that add a missed method for the Domain builder.


pvoborni (approved):
LGTM

PR #357, opened 2024-10-16
refactor(HMS-3803): helperTestFindByIDIpa to FindIpaByID

avisiedo:
This change refactor helperTestFindByIDIpa function
into a simpler way, starting to create the package
internal/test/sql/ipas which will hold the code for
the SQL mock helpers, and making them more simple.

The public functions will match with the real
method/function. That helper will be currently
orquestating the preparation of the sql mock
statement. The first 3 arguments would be
(stage, mock, expectedErr, ...) where stage
is the step in the control flow, mock is the
sqlmock instance where prepare the expectations,
and expectedErr is nil for success or an error
reference for error expectations.

Additional functions named PrepSql... will
mock the specific SQL statement, decomposing
complex executions in smaller units. This functions
will haveas 3 first arguments: mock, withError and
expectedErr
- mock is the sqlmock.Sqlmock instance which will
  receive the expectations.
- withError will be the result of the helper
  WithPredicateExpectedError(step, stage, expectedErr)
- expectedErr is nil (when no error expected) or the
  error to be returned.

The predicate WithPredicateExpectedError simplify to
think about when the error should be returned, and remove
duplications on the logic as it always check if the current
step is the final (stage) and if an error expectations is
provided, it is prepared to be returned on that stage. This
helps to get the error returned in the specific step.

### DIFF
--- a/internal/test/sql/helper.go
+++ b/internal/test/sql/helper.go
@@ -1,0 +1,6 @@
+package sql
+
+// WithPredicateExpectedError
+func WithPredicateExpectedError(step, stage int, expectedErr error) bool {
+	return step == stage && expectedErr != nil
+}
--- a/internal/test/sql/ipas_sql.go
+++ b/internal/test/sql/ipas_sql.go
@@ -1,0 +1,147 @@
+package sql
+
+import (
+	"fmt"
+	"regexp"
+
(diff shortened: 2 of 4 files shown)


pvoborni (approved):
LGTM

PR #383, opened 2024-10-24
Move to Konflux image

mshriver:


deployments/clowder.yaml
```diff
@@ -204,7 +204,7 @@ parameters:
       The internal application name that is used to
       compose the API path for its routes.
   - name: IMAGE
-    value: quay.io/cloudservices/idmsvc-backend
+    value: redhat-services-prod/insights-management-tenant/insights-idmsvc/idmsvc-backend
```
mshriver, line 207: this is the one that matters for deployment

mshriver:
PR check is failing, app-interface change to add IMAGE param for ephemeral will fix it - and we'll also be moving away from it very soon.

PR #385, opened 2024-10-24
Fix image path, quay.io was missed

mshriver:


avisiedo (approved):
LGTM thanks for the fix

PR #394, opened 2024-10-31
fix(HMS-4922): fix grafana dashboard

avisiedo:
This change add a variable to the dashboard which
represent the datasource and it will point out to
the prometheus instance.

https://issues.redhat.com/browse/HMS-4922


pvoborni (approved):
LGTM

PR #395, opened 2024-10-31
feat(HMS-4931): enhance grafana configmap generation

avisiedo:
This change add a script to generate the grafana configmap and split the grafana dashboard in an independent json file.

TODO:
- Keep the header of the template as it is, to keep the data section at the end.
- Write the content as string with \n representation.
- Update github to check we don't miss to run the script.


frasertweedale:
Closing as won't-do.

PR #428, opened 2024-11-29
fix(HMS-5066): make config.example.yaml secure by default

frasertweedale:
Set `app.accept_x_rh_fake_identity: false` in the example config
file (which is copied as-is to the container image).  In this way,
the default configuration will be safe.

Add a comment that this file should have safe/appropriate values for
production use.  Overrides, when necessary, should be effected via
make variables, environiment variables, or modifying the config.

Fixes: https://issues.redhat.com/browse/HMS-5066


avisiedo (changes requested):
Thank you for the change! I agree with it.

I have reviewed additional files and I think the below could be required at `deployments/backend-compose.yaml`:

```yaml
backend-service:
  environment:
    - APP_ACCEPT_X_RH_FAKE_IDENTITY=true
```

Tracking the file: it is referenced from [ipa-hcc](https://github.com/podengo-project/ipa-hcc/tree/main) repository at 
- [idm-ci/playbooks/idm-domains-backend-deploy.yaml](https://github.com/podengo-project/ipa-hcc/blob/main/idm-ci/playbooks/idm-domains-backend-deploy.yaml#L8).
  - [hmsidm-rhel93-backend.yaml](https://github.com/podengo-project/ipa-hcc/blob/main/idm-ci/metadata/hmsidm-rhel93-backend.yaml#L54).
  - [hmsidm-domains-backend.yaml](https://github.com/podengo-project/ipa-hcc/blob/main/idm-ci/metadata/hmsidm-domains-backend.yaml#L47).

After that it is a LGTM.

frasertweedale:
Thanks, updated and rebased.  I will merge when CI passes.

